PR #10, opened 2018-04-18
fix(Update toRoundedUnit to properly round negative values)

### LINKED ISSUE #7: Negative numbers rounded incorrectly
Math.round rounds the wrong way for negative numbers:

Dinero({amount: 205}).toRoundedUnit(1) // 2.1 (expected 2.1) (correct)
Dinero({amount: -205}).toRoundedUnit(1) // -2 (expected -2.1) (wrong)

tarnelope:
Math.round() doesn't correctly round negative values so this change first rounds the absolute value of the given amount and then multiplies in the Math.sign() afterwards.

This is my first contribution so happy for any feedback and change requests.

fix #7

### DIFF
--- a/src/dinero.js
+++ b/src/dinero.js
@@ -574,8 +574,12 @@ const Dinero = options => {
      */
     toRoundedUnit(precision) {
       const factor = Math.pow(10, precision)
+
       return calculator.divide(
-        Math.round(calculator.multiply(this.toUnit(), factor)),
+        calculator.multiply(
+          Math.sign(this.toUnit()),
+          Math.round(Math.abs(calculator.multiply(this.toUnit(), factor)))
+        ),
         factor
       )
     },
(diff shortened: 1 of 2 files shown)


tarnelope:
Sorry about that! I noticed those afterwards and wasn't sure if I could edit the message but I think I figured it out.

src/dinero.js
```diff
@@ -574,8 +574,10 @@ const Dinero = options => {
      */
     toRoundedUnit(precision) {
       const factor = Math.pow(10, precision)
+
       return calculator.divide(
-        Math.round(calculator.multiply(this.toUnit(), factor)),
+        Math.sign(this.toUnit()) *
```
sarahdayan: I was just thinking this should be done with `Calculator#multiply` as well. Will be more future-proof when support for big integers is added.

```
calculator.multiply(Math.sign(this.toUnit()), Math.round(Math.abs(calculator.multiply(this.toUnit(), factor))))
```

Issue #13, opened 2018-04-20
Allow specifying the precision?


scotttrinh:
Any plan or desire to allow specifying the precision with `2` being the default? Seems useful to account for regulatory requirements or even just personal preference. I'm imagining something like this:

```js
Dinero({ amount: 1000, currency: 'USD' })
  .toFormat('$0,0.00'); // '$10.00'
Dinero({ amount: 10000000, precision: 6, currency: 'USD' })
  .toFormat('$0,0.00'); // '$10.00'
```

I'd be happy to PR this!

sarahdayan:
There's a [project](https://github.com/sarahdayan/dinero.js/projects/3) for this. You can click "Menu" to see the description. I'm assigning it to this issue.

I feel like this deserves a discussion before starting to code 😊

scotttrinh:
Awesome. Do you have an idea about how to have this discussion? Just talk about it on this issue or set up a call?

I can throw together a throw-away PR that has the most naivë implementation I can think of as a starting point for a discussion if that helps materialize some ideas and concerns.

sarahdayan:
There are two approaches for adding the concept of precision: **added to**, or **baked in** the [exponent](https://en.wikipedia.org/wiki/ISO_4217#Treatment_of_minor_currency_units_(the_%22exponent%22)).

_For the record, the exponent is the factor by which you multiply 10 to get the number of sub-units in a unit. For dollars, the exponent is `2` as `10^2 = 100`._

Dinero.js doesn't support exponents properly yet, [but will in v2](https://github.com/sarahdayan/dinero.js/projects/5). Now even though it doesn't currently have the concept of an exponent, it implicitly does by dividing by `100`. This implicit behavior was reflected in [v1.1.2](https://github.com/sarahdayan/dinero.js/releases/tag/v1.1.2) to prepare for future support.

The exponent defacto indicates the default number of possible decimal places for a _billable_ amount. Yet you may want to represent prices with more precision:

- unit price for items sold in higher quantity, like stuff in bulk or gas
- avoid rounding during calculations to preserve precision

So this concept, loosely named precision here, aims at "adding" support for these fractional units so they can be represented and calculated.

Back to the two approaches, here how you can think about it:

### Solution 1: Precision is added to the exponent

In this case, precision starts at 0. Anything more adds up to the existing exponent of the currency.
Whatever the exponent is, precision adds up decimal places.

```js
Dinero({ amount: 104545, currency 'USD' })
[...]

PR #14, opened 2018-04-20
feat(dinero): Allow specifying a precision to store internal value

scotttrinh:
Totally naive implementation for storing the value at a higher precision. A starting point for discussion.

This immediately jumped out to me: The value and precision being stored in a single `Number` has limitations on the size of the final number we can represent. I've limited the precision to 15 here, but it's so easy to approach the `Number.MAX_SAFE_INTEGER` here with a few multiplication operations. The current implementation suffers from this, too, but it's more likely to happen at higher precision values than at the current value of `2`.

re #13

### DIFF
--- a/src/dinero.js
+++ b/src/dinero.js
@@ -59,19 +59,26 @@ const Dinero = options => {
       if (!Number.isInteger(number)) {
         throw new TypeError('You must provide an integer.')
       }
+    },
+    isValidPrecision(number) {
+      if (number < 1 || number > 15) {
+        throw new TypeError('You must provide an integer between 1 and 15.')
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


scotttrinh:
A few things I've thought about over the weekend:

- It'd be nice to be able to "Bring Your Own BigInt" to overcome the max number issue.
- No need to pick for the user and bloat this library, but give some affordances for using a library like `big-integer`.
- Users would need to provide their own `calculator`.
- Default to `Number` of course.

sarahdayan:
After the conversation on #13, I'll close this PR and open a new one. @scotttrinh I'll tackle this issue since you volunteered to start working on the BigInt integration 👍

PR #17, opened 2018-04-27
Add convert method

sarahdayan:
This PR adds a `convert` method to `Dinero`, allowing for users to convert Dinero objects from one currency to another.

Users need to plug their own forex API to retrieve exchange rates. The `convert` method queries a `Forex` service which calls the API and returns a promise with the right exchange rate for the destination currency. Then, `convert` creates a new Dinero object with the converted amount and new currency, and returns it within a promise.


src/services/helpers.js
```diff
@@ -82,3 +82,83 @@ export function countFractionDigits(number = 0) {
...
+      onerror() {
+        reject(new Error('Network error'))
+      }
+    })
+
+    request.open('GET', url, true)
+    setXHRHeaders(request, options.headers)
+    request.send()
+  })
+}
```
scotttrinh: Not sure if there is an explicit browser-support contract, but any reason not to use `fetch` here? Seems like any browser that supports `Object#assign` would also support `fetch`.

sarahdayan: I initially used it but when it was an absolute pain when it came to testing it with Jest, and I couldn't find a way to do it without yet another dev dependency. I may be wrong though, would be happy to know if there's a simpler solution.

src/services/helpers.js
```diff
@@ -82,3 +82,83 @@ export function countFractionDigits(number = 0) {
...
+    key = encodeURIComponent(key)
+    return [
+      string,
+      i === 0 ? '?' : '&',
+      key,
+      '=',
+      encodeURIComponent(object[key])
+    ].join('')
+  }, '')
+}
```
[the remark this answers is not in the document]
sarahdayan: Good call! I didn't know that object. If that's _polyfillable_ with Babel I'm all for it. Going to look into it.

src/services/helpers.js
sarahdayan: That's what I thought, `Fetch` needs [Babel Polyfill](https://babeljs.io/docs/usage/polyfill/) as a **dependency** (and not a dev dependency) to be made backward-compatible with IE11. Since this is a library and not an app, I'm a bit reluctant to adding dependencies unless this is absolutely necessary.

src/services/helpers.js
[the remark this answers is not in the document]
sarahdayan: Same as `Fetch`.

src/services/helpers.js
sarahdayan: I'd like to be backward-compatible at least down to IE11 (I shipped polyfilled versions today in [v1.2.0](https://github.com/sarahdayan/dinero.js/releases/tag/v1.2.0) as I realized there were some holes). I picked [core-js](https://github.com/zloirock/core-js) for polyfills (the same that Babel uses), and unfortunately, it doesn't provide one for `Fetch`.

I want to avoid adding dependencies for each feature, so let's wait for this one and refactor later when it's time to drop IE altogether.

src/services/helpers.js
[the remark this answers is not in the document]
sarahdayan: URL isn't polyfilled in core-js [yet](https://github.com/zloirock/core-js/issues/117), let's use it when it is.

sarahdayan:
@scotttrinh (and/or anyone else) Refactored a few things and made the method more API-proof. Would love a final review before merge if possible 🙂

src/services/settings.js
```diff
@@ -28,16 +28,33 @@ export const Defaults = {
  * @property {String}  globalFormat - The global format for new Dinero objects (see {@link module:Dinero~toFormat toFormat} for format).
  * @property {String}  globalRoundingMode - The global rounding mode for new Dinero objects (see {@link module:Dinero~multiply multiply} or {@link module:Dinero~divide divide} for format).
  * @property {String}  globalFormatRoundingMode - The global rounding mode to format new Dinero objects (see {@link module:Dinero~toFormat toFormat} or {@link module:Dinero~toRoundedUnit toRoundedUnit} for format).
+ * @property {String}  globalExchangeRatesApi.endpoint - The global exchange rates API endpoint for new Dinero objects (see {@link module:Dinero~convert convert} for format).
+ * @property {String}  globalExchangeRatesApi.JSONPath - The global exchange rates API JSON path for new Dinero objects (see {@link module:Dinero~convert convert} for format).
```
scotttrinh: Do you think it might be a bit misleading to call this `JSONPath` since it doesn't seem to actually be [JSONPath](http://goessner.net/articles/JsonPath/index.html#e2)?

[Lodash has a similar functionality](https://lodash.com/docs/4.17.10#property) which takes arrays into account which we might consider using here so that there is a well-understood/tested interface.

scotttrinh: I did some testing and you can access arrays by using the index `'b.c.1'`, so that works for me, but I'm still not sure we should call it JSONPath just because of the ambiguity.

sarahdayan: You're totally right about the name ambiguity. I'll go for `propertyPath`.

sarahdayan:
@scotttrinh Cool! Merging it now, deploying it probably later next week. Thanks for the review!

Issue #9, opened 2018-04-17
Support ISO 4217 minor currency units (cents vs fils)


montanaflynn:
If you want the library to support all ISO 4217 currencies as suggested in the documentation you may want to rethink using "cents" as the amount value. For instance some currencies do not have a [minor currency equivalent](https://en.wikipedia.org/wiki/ISO_4217#Treatment_of_minor_currency_units_(the_%22exponent%22))  (e.g. Japanese Yen) and others have different exponents (e.g. The Iraqi Dinar is divided into 1000 fils. The United States Dollar is divided into 100 cents.) so maybe that should be handled by the library. You can find all the [currencies and their minor currency exponent here](https://en.wikipedia.org/wiki/ISO_4217#Active_codes).

```js
// this should be 1 Iraqi Dinar not 10
Dinero({ amount: 1000, currency: 'IQD' })
```

sarahdayan:
Thank you @montanaflynn for raising this issue.

I'm definitely considering offering better support for all ISO 4217 currencies. Instead of revising the usage of sub-units as the amount (which covers many cases) there could be something to do with the exponent. For example, you could indicate it as you create a Dinero object and it would become the divisor for every manipulation.

```
Dinero({ amount: 1000, currency: 'IQD', exponent: 3 })
```

One step further would be creating a Currency service, which would allow currency assertions and automatic exponent inclusion. This would need to be maintained though.

pawel-buczkowski-payu:
Your library is really cool!

But I really need this feature described in this issue. When do you plan to release v2?

BTW. It would be also nice to add a similar feature in toFormat function: when I use this function ex. toFormat('$0,0.00') I have to type so many zeros (at the end, after dot) as the currency precision is. It would be nice to have a way to define format like: toFormat('$0,0.#'), so the script will convert it to appropriate number of digits depending on currency precision. So for number 1234 it would give results: 12.34 for EUR, 1,234 for JPY and 1.234 for IQD :)

andyburke:
Although it is uncommon (I only see few in Wikipedia's list of circulating currencies: https://en.wikipedia.org/wiki/List_of_circulating_currencies ), some currencies are not base-10:

```
  |--------------------|----------|---------|----------|----------------|-----------------|
  | State or territory | Currency | Symbol  | ISO code | Fractionalunit | Number to basic |
...
  | Madagascar         | ...      | Ar      | MGA      | Iraimbilanja   | 5               |
```

[...]

PR #19, opened 2018-05-04
Add precision concept

sarahdayan:
This PR adds the `precision` concept to the library (as discussed in #13). This allows user to specify how many decimal places their amount represents, so they can have more precise Dinero objects. This also provides a way to circumvent the problem raised in #9 (but this will be fixed in v2).

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ Dinero({ amount: 10545, precision: 3 })
 Dinero({ amount: 1, currency: 'JPY', precision: 0 })
 ```
 
-Of course, if you're using the same currency more than once, it might be worth setting a default precision.
+If you're using the same currency more than once, it might be worth setting a default precision.
 
 ```js
 // The Iraqi dinar has up to 3 sub-units
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Dinero.js makes it easy to create, calculate and format monetary values in JavaS
 
 **Note:** The library is globally available in the docs for you to be able to test it right in the browser console.
 
-To get started, you need to create a new Dinero instance. Amounts are specified in **cents**. You can also specify an [ISO 4217 currency code][wiki:iso-4217] (default is `USD`).
+To get started, you need to create a new Dinero instance. Amounts are specified in **minor currency units** (e.g.: "cents" for the dollar). You can also specify an [ISO 4217 currency code][wiki:iso-4217] (default is `USD`).
 
 This represents €50:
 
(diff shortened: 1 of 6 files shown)


src/dinero.js
```diff
@@ -343,14 +410,21 @@ const Dinero = options => {
...
+     * Dinero({ amount: 800 }).lessThan(Dinero({ amount: 5000, precision: 3 }))
      *
      * @throws {TypeError} If `comparator` has a different currency.
      *
      * @return {Boolean}
      */
     lessThan(comparator) {
       assertSameCurrency.call(this, comparator)
-      return this.getAmount() < comparator.getAmount()
+      const comparators = Dinero.normalizePrecision([this, comparator])
```
zellwk: I'm not sure `normalizePrecision` is a good method name. While reading this code, I didn't understand what it would do, and I had to download the source to check for it. 

From your comments, it seems like you're setting to the highest precision, so maybe a better name will be `setToHighestPrecision` or something similar?

sarahdayan: I hear you, but I feel like "normalize" is a pretty common term in programming when it comes to making elements consistent, unified, "normalized". For example, `normalize.css` makes browsers render all elements more consistently.

`setToHighestPrecision` is more accurate according to what the method *does*, but it's not it's *purpose*. It's a good name for an internal method, I don't think it is for a public API (or there would need to be a `toHighestPrecision` and `toLowestPrecision`).

Does that make sense?

README.md
```diff
@@ -157,6 +157,28 @@ Dinero({ amount: 500 })
...
+```js
+// represents $10.4545
+Dinero({ amount: 10545, precision: 3 })
+
+// The Japanese yen doesn't have sub-units
+// this represents ¥1
+Dinero({ amount: 1, currency: 'JPY', precision: 0 })
+```
+
+Of course, if you're using the same currency more than once, it might be worth setting a default precision.
```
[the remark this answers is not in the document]
sarahdayan: Good catch!

src/dinero.js
zellwk: @sarahdayan I hear you, and it makes sense. Just wondered if there's a better alternative. I'll keep my brain open for the next few days and let you know again if something comes up.

zellwk: How about `normalizeToHighestPrecusuon`? 

The name is slightly longer, but it speaks so much more about the function.

sarahdayan: It's the same problem. If you specify "to highest" then you're giving a clue about how it works under the hood instead of focusing on the purpose. You can't normalize in any other way without losing information, so there's no reason to have a `normalizeToLowestPrecision` method. Thus, giving this extra "to highest" precision doesn't seem necessary. That's what the docs are here for.

I'm all for self-explaining method names, but there needs to be a balance. Overly lengthy names that can be shortened thanks to good documentation without losing the gist of what it's doing are a sweet spot IMO.

zellwk: Sure thing. It's your call :)

Issue #22, opened 2018-06-08
Default serialization output for Dinero with toJSON


hanxue:
## Expected behavior
The output of  `JSON.stringify(Dinero())` should be the same as `Dinero.toFormat()` with it's default or global values. For example: `$5,500.00`

## Actual behavior
Currently the output of `JSON.stringify(Dinero())` is `{}`, which is not useful. 

## Proposal
Implement `toJSON()` in Dinero so that Dinero objects can be serialized when calling `JSON.stringify()`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior

The intent here is to make Dinero "just works", for the same reason the following Global settings are pre-defined

```javascript
  globalLocale: 'en-US',
  globalFormat: '$0,0.00',
  globalRoundingMode: 'HALF_EVEN',
  globalFormatRoundingMode: 'HALF_AWAY_FROM_ZERO',
```

sarahdayan:
Hi @hanxue, thanks for opening this issue and sorry for the delayed reply.

I'm curious, why do you think `JSON.stringify(Dinero())` should have the same output as `Dinero.toFormat()`? I'm not sure I understand the logic of it, nor the usefulness:

1. `JSON.stringify` is supposed to turn a JavaScript value into valid JSON. Currently, it outputs `{}` because a Dinero object contains functions, not static values. But to me, if we were to expect a different behavior, I would imagine something closer to what you'd get if you did `JSON.stringify(Dinero().toObject())` (that is, a JSON representation of a given Dinero object with its main properties).
2. Why create a new method that does the same thing as an existing one (in your case `toFormat`)?

I'm potentially open to a `toJSON` method, but there needs to be a benefit for users. Otherwise, we're just adding bloat.

Closing this, but open to discussion 🙂

yacinehmito:
> I'm potentially open to a toJSON method, but there needs to be a benefit for users. Otherwise, we're just adding bloat.

Would you be open to have `toJSON = toObject`?

At Spendesk we're experimenting with Dinero. Down the line we might have a single representation for all our monetary values and this representation must be carried on the wire, wherever it might me.

We can to it without issue on our end, but I think that would be a sensible default behaviour (currently `JSON.stringify(Dinero({ amount: 0, currency: 'EUR' }))` outputs `{}`).

sarahdayan:
[...]

PR #23, opened 2018-06-08
feat(dinero.js): implement toJSON() to serialize output of Dinero object

hanxue:
Implement toJSON() so that calls to JSON.stringify(Dinero()) will return a string. See #22

#22


sarahdayan:
Thanks @hanxue for working on this, but I'm not going to merge for the reasons I brought up in #22.

Issue #24, opened 2018-06-09
Missing Typescript declaration file


zach-albia:
I think a Typescript declaration file would be a huge help for this library's users, JS and TS alike.

sarahdayan:
Hi Zach! You're not the first person asking for TypeScript support, and I'm open to this. A few things on that topic:

1. I'm not myself familiar with TypeScript. I could study it, but this isn't my focus at the moment, and it would take time if I were to do it myself. **Help is welcome**.
2. I feel like it would be great to create a separate repo for this. I own the `dinerojs` Github organization and am thinking of moving Dinero.js under that account at some point, and I feel like it could be good to create several repositories for each usage. This way, you don't bloat the JS package for those who aren't interested in TS, and vice-versa.

Leaving this open for now in case someone wants to get more involved in helping for TypeScript support. Let's chat.

Bessonov:
Hey, @sarahdayan I'm interested in TS support too.

From my point of view, and how many other projects do that, is best to use TypeScript as first development language and compile the output to JavaScript. You get robust, type safe and nice code.

In any way, support of TypeScript doesn't bloat the JS package. You can stay with JS and just append TS typings, which are downloaded with npm or yarn, but they are used for development only and not delivered to the clients.

BirdInTheCity:
I, too, am hoping for TS support. Not sure what's required to make the project TS complaint. I've been having to use type `any` instead of `Dinero`, which isn't ideal.

BirdInTheCity:
For anyone looking to pick up the task, here's how to do it:  
https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

I'm already slammed at work, so I don't think I'll have time in the coming weeks. Perhaps after that though?

Bessonov:
DefinitelyTyped is for external typings. For best TS support is [1. bundling with your npm package](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) the way to go.

Bessonov:
> Not sure what's required to make the project TS complaint.
AFAIK, there are two primary ways. Port everything to typescript (best from my point of view) or create typings.

zach-albia:
Hi all, I'm pretty new to TypeScript myself. I made an attempt at types for dinero some two weeks ago. Please see [this gist](https://gist.github.com/zach-albia/0e4ec43351dd594447ebe3703de631f2). Any input would be appreciated.

BirdInTheCity:
[...]

PR #25, opened 2018-07-13
Added types

zach-albia:
I'm opening this PR as per [issue #24](https://github.com/sarahdayan/dinero.js/issues/24). I understand that adding the types file to the project may not be to your liking. If so, please advise if you would prefer them to be outside the repo instead, e.g. in DefinitelyTyped.

### DIFF
--- a/src/dinero.d.ts
+++ b/src/dinero.d.ts
@@ -1,0 +1,88 @@
+declare namespace dinero {
+  type Options = Partial<ObjectLiteral>
+
+  interface ConversionOptions {
+    endpoint: string,
+    propertyPath: string,
+    headers?: object,
+    roundingMode?: RoundingMode
+  }
+
+  type RoundingMode = 'HALF_ODD'
+    | 'HALF_EVEN'
+    | 'HALF_UP'
+    | 'HALF_DOWN'
+    | 'HALF_TOWARDS_ZERO'
+    | 'HALF_AWAY_FROM_ZERO'
+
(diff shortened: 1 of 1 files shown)


Bessonov:
Hey @zach-albia, I miss `types` entry in package.json

src/dinero.d.ts
```diff
@@ -0,0 +1,88 @@
...
+  hasSameAmount(comparator: Dinero): boolean;
+  toFormat(format?: string, roundingMode?: dinero.RoundingMode): string;
+  toUnit(): number;
+  toRoundedUnit(digits, roundingMode?: dinero.RoundingMode): number;
+  toObject(): dinero.ObjectLiteral;
+  normalizePrecision(objects: Dinero[]): Dinero[];
+}
+
+declare module 'dinero.js' {
+  export = Constructor
```
BirdInTheCity: I'm getting errors about needing a default export. It worked for me when changed to:
`export default Constructor`

zach-albia: Hey, thanks for catching that. will update pull request accordingly.

zach-albia:
I'm looking to add the following to package.json line 15 : `"types": "./src/dinero.d.ts",` as per comment by @Bessonov. I don't know what line 14 `"main": "build/cjs/dinero.js",` does but it looks like it does something related to CommonJS.

Should I instead set the `types` attribute to something like `"types": "build/cjs/dinero.d.ts",`?

PR #29, opened 2018-08-16
Added typings and updated build

zach-albia:
Hi, made some time to do #24 right. Please advise if changes to `tasks/build.js` are okay. I could have opted to run the copy _after_ all the promises resolve but it would involve more changes to the build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "homepage": "https://sarahdayan.github.io/dinero.js",
   "repository": "https://github.com/sarahdayan/dinero.js",
   "main": "build/cjs/dinero.js",
+  "types": "build/ems/dinero.d.ts",
   "scripts": {
     "precommit": "lint-staged",
     "cz": "git-cz",
--- a/src/dinero.d.ts
+++ b/src/dinero.d.ts
@@ -1,0 +1,81 @@
+export type Options = Partial<ObjectLiteral>
+
+export interface ConversionOptions {
+  readonly endpoint: string
+  readonly propertyPath: string
+  readonly headers?: object
+  readonly roundingMode?: RoundingMode
+}
+
+// Omit taken from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+type ExchangeRatesApi = Omit<ConversionOptions, "roundingMode">
+
+export type RoundingMode =
+  | 'HALF_ODD'
+  | 'HALF_EVEN'
+  | 'HALF_UP'
+  | 'HALF_DOWN'
+  | 'HALF_TOWARDS_ZERO'
+  | 'HALF_AWAY_FROM_ZERO'
+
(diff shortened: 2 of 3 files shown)


BirdInTheCity:
@sarahdayan  Any updates on when this will be merged? It seems like the failing test is unrelated to this PR.

sarahdayan:
@BirdInTheCity @zach-albia @itchingpixels Hi! Sorry for the delay.

I've fixed the faulty build, the PR should pass now. I'll look into it as soon as I get the chance, will do my best to do it as soon as possible.

burabure:
@sarahdayan I' currently shimming these typings into my TS project, but I'd love to have this "natively" on the npm package; Let me know if I can help you with getting this merged =)

sarahdayan:
Hi! Thanks all for your involvement and bearing with me. I think we’re close to getting this shipped 🤩

The TypeScript file looks good to me, but I see two things that should be fixed before merging:

1. After asking around, it seems that the recommendation is to generate one .d.ts per published .js file. In this case, this would mean 4 files per module definition (`dinero.d.ts`, `dinero.min.d.ts`, `dinero.polyfilled.d.ts`, `dinero.polyfilled.min.d.ts`), so in total 16 copies (one for `esm/ˋ, one for ˋcjs/ˋ, one for ˋamd/ˋ, one for ˋumd/ˋ). The `types` entry in the `package.json` should remain the same though.
2. The file copying should be asynchronous so it doesn’t make the build process longer. Rollup’s ˋbundle.write` likely returns a promise (to be checked) so copying files asynchronously after each .js build is generated in `buildOutputs` would Best regards cleaner.

Tell me if you need precisions.

devzsolt:
@zach-albia we would love to see your PR merged, could you please take another look at the suggestions of @sarahdayan ?

zach-albia:
Hi @devzsolt, sorry was busy with a project. Now that I have some free time in between, I can get this done probably within the week.

zach-albia:
@sarahdayan or anyone else, would you happen to have an example of a project that follows the pattern you mentioned in the following quote? I'm not quite sure how to process `dinero.d.ts` and what the output of that process should be.

> 1. After asking around, it seems that the recommendation is to generate one .d.ts per published .js file. In this case, this would mean 4 files per module definition (`dinero.d.ts`, `dinero.min.d.ts`, `dinero.polyfilled.d.ts`, `dinero.polyfilled.min.d.ts`), so in total 16 copies (one for `esm/`, one for `cjs/`, one for `amd/`, one for `umd/`). The `types` entry in the `package.json` should remain the same though.

sarahdayan:
Looks like [someone created a definition file](https://www.npmjs.com/package/@types/dinero.js) on DefinitelyTyped in the meantime. Sorry for not making this move fast enough @zach-albia but it looks like this solves the issue altogether, so, still a good thing!

I'll close this PR now, please PR on [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) if there's an issue with the definition file they have.

zach-albia:
It's all good.

PR #39, opened 2018-12-17
Update to fix provided command to build dist files

frobinsonj:
Replaced `npm build` with `npm run build`. It previously didn't work due to the existing [npm build command](https://docs.npmjs.com/cli/build.html).

Sorry if I have gone about this the wrong way, hopefully this is a welcome change! :)


frobinsonj:
Just realised commit should be `docs(contributing): Fix provided command to build dist files`. I will create a new commit and pull request, my bad,

PR #85, opened 2020-02-02
fix(dinero#allocate): do not throw exception when some ratio is 0

dotpack:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Compliance

- [x] My change isn't breaking (it doesn't cause existing functionality to change).
- [x] I have read the [CONTRIBUTING](https://github.com/sarahdayan/dinero.js/blob/develop/CONTRIBUTING.md) document.
- [x] My code follows the coding style of this project.
- [x] I have properly formatted my commit messages with [cz-cli](https://github.com/commitizen/cz-cli), or manually, following the [Angular Commit Messages Guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines).
- [x] I have updated the documentation accordingly, or my changes doesn't require documentation changes.
- [x] I have added tests to cover my changes, or my changes doesn't require new tests.
- [x] All new and existing tests pass.


sarahdayan:
Hi there! Why would you need to be able to allocate to a `0` ratio?

dotpack:
Hello @sarahdayan. Thank you for you response. In my case I have a lot of work with prices where one or more of it can be a `0`.

Every time when I need to allocate bonus between my prices I have to filter price items and then concatenate allocation result in the correct order.

In the current version I have to write custom allocate function like this:

```
function withZeroAllocate(dinero, rates) {
	const nonZeroRates = rates.filter(Boolean);
	const allocation = dinero.allocate(nonZeroRates);
	const resultAllocation = [];
	let index = 0;
	for (let i = 0; i < rates.length; i += 1) {
		if (rates[i]) {
			resultAllocation.push(allocation[index].toUnit());
			index++;
		} else {
			resultAllocation.push(rates[i])
		}
	}
	return resultAllocation;
}
```

I think the ability to allocate to a `0` ratio is mathematical correct and will be very useful. `js-money` package allows it.

test/unit/dinero.spec.js
```diff
@@ -247,8 +247,11 @@ describe('Dinero', () => {
       expect(shares[0].getAmount()).toBe(25)
       expect(shares[1].getAmount()).toBe(75)
     })
-    test('should throw when one of the ratios equals to zero', () => {
-      expect(() => Dinero({ amount: 1003 }).allocate([60, 0, 10, 30])).toThrow()
+    test('should allocate the amount of the Dinero object into new ones and distribute the remainder when one of the ratios equals to zero', () => {
+      const shares = Dinero({ amount: 1003 }).allocate([50, 0, 50])
+      expect(shares[0].getAmount()).toBe(502)
+      expect(shares[1].getAmount()).toBe(0)
+      expect(shares[2].getAmount()).toBe(501)
```
sarahdayan: This use case works because you're splitting amounts in 2 (50/50) and because the first ratio is > 0. The remainder is 1, therefore it's distributed to the first ratio and it stops there.

If you do `Dinero({ amount: 1003 }).allocate([0, 50, 50])`, this will fail because the remainder will distribute to the first ratio (0) without verifying if it should.

I'd recommend adding that test case (or using the one that you deleted) and to adapt the algorithm of the `allocate` method to make it pass.

Additionally, I realized that the way I used to assert for `allocate` (several assertions in a single test) isn't optimal, because Jest exits on the very first breaking test and doesn't show you the whole distributed ratios.

Something like this would work better IMHO (I'll refactor the other tests in another PR):

```js
expect(shares.map(share => share.getAmount())).toEqual([0, 502, 501])
```

Jest feedback:

Capture d’écran 2020-03-07 à 13 32 00

dotpack: Ouch. I realized that the better way to allocate remainder would be allocate it to ordered ratios, but it will be breaking change. Right now I've saved remainder allocation order.

I've added the test to illustrate my idea.
```js
const shares = Dinero({ amount: 1 }).allocate([0, 1, 50])
expect(shares.map(share => share.getAmount())).toEqual([0, 1, 0])
```

I think in the `2.0.0` version the order of the result items should be changed.

Thank you.

dotpack: Additionally, could you please fix this line:
https://github.com/dinerojs/dinero.js/blob/develop/.all-contributorsrc#L3

When I run `npm run contributors:generate` all contributor tags was changed to `https://github.com/sarahdayan/dinero.js`

test/unit/dinero.spec.js
```diff
@@ -239,16 +239,22 @@ describe('Dinero', () => {
...
-      expect(() => Dinero({ amount: 1003 }).allocate([60, 0, 10, 30])).toThrow()
+    test('should allocate the amount of the Dinero object into new ones and distribute the remainder when one of the ratios equals to zero', () => {
+      const shares = Dinero({ amount: 1003 }).allocate([0, 50, 50])
+      expect(shares.map(share => share.getAmount())).toEqual([0, 502, 501])
+    })
+    test('should allocate the small amount of the Dinero object into new ones and distribute the remainder when one of the ratios equals to zero', () => {
+      const shares = Dinero({ amount: 1 }).allocate([0, 1, 50])
+      expect(shares.map(share => share.getAmount())).toEqual([0, 1, 0])
+    })
+    test('should throw when all of the ratios is zero', () => {
```
sarahdayan:
```suggestion
    test('should throw when all ratios are equal to zero', () => {
```

test/unit/dinero.spec.js
```diff
@@ -239,16 +239,22 @@ describe('Dinero', () => {
...
     })
     test('should allocate the amount of the Dinero object into new ones and distribute the remainder when given ratios', () => {
       const shares = Dinero({ amount: 100 }).allocate([1, 3])
-      expect(shares[0].getAmount()).toBe(25)
-      expect(shares[1].getAmount()).toBe(75)
+      expect(shares.map(share => share.getAmount())).toEqual([25, 75])
     })
-    test('should throw when one of the ratios equals to zero', () => {
-      expect(() => Dinero({ amount: 1003 }).allocate([60, 0, 10, 30])).toThrow()
+    test('should allocate the amount of the Dinero object into new ones and distribute the remainder when one of the ratios equals to zero', () => {
```
sarahdayan:
```suggestion
    test('should not distribute the remainder to zero ratios', () => {
```

test/unit/dinero.spec.js
```diff
@@ -239,16 +239,22 @@ describe('Dinero', () => {
...
+      const shares = Dinero({ amount: 1003 }).allocate([0, 50, 50])
+      expect(shares.map(share => share.getAmount())).toEqual([0, 502, 501])
+    })
+    test('should allocate the small amount of the Dinero object into new ones and distribute the remainder when one of the ratios equals to zero', () => {
```
sarahdayan: It's a bit unclear to me what this test demonstrates. Can you please clarify the description?

dotpack: Sorry, I should have described the problem more clearly.

I deleted the test. Probably we should talk about it in another issue.

The test shows a problem with the distribution of the remainder between the ratios `[1, 99]`. The current version returns the distribution of `[1, 0]`. Probably next major version should returns the distribution of `[0, 1]`.

sarahdayan: No worries. I addressed this point [here](https://github.com/dinerojs/dinero.js/pull/85#pullrequestreview-371924185) but indeed, feel free to open a new issue so we can elaborate more on it.

Issue #73, opened 2019-11-07
Not async call .convert() for own provided rates


WojtylaCZ:
**Is your feature request related to a problem? Please describe.**

I would find very convenient to stay in synchronous context when using the convert() method with own exchange rates. Without using a immediately resolving Promise.

**Describe the solution you'd like**

Add optional parameter to convert method for object with rates. It will work with CurrencyConverter as the Promise is only for API calls.

**Describe alternatives you've considered**

Manual conversion with multiplication and creating new Dinero object.

**Additional context**

Relates to #17 

Thank you for your work and the package. Great job! :)

baskin:
Same requirement. In my case I use mobx managed exchange rates in a react environment. I'll be forced to run side effects (useEffect hooks etc..) to resolve a promise which I know resolves locally.

sarahdayan:
Hey! I've just released [Dinero.js v2 alpha](https://v2.dinerojs.com/docs) with a much better design for conversions. You can provide rates directly, the responsibility of fetching them isn't coupled to the library.

You can check it out in the [API reference](https://v2.dinerojs.com/docs/api/conversions/convert).

PR #206, opened 2021-03-18
feat(dinero): add synchronous convert method to API

vipera:
Add an alternative, synchronous API to the async convert() that can be used synchronously when the
exchange rates are known at the moment of conversion.

re #73

## Types of changes

As discussed in #73, this PR adds a synchronous version of the convert function for use cases where the exchange rate table is immediately available, and does not need to possibly be fetched from an API.

The project I'm working on also uses locally cached exchange rates retrieved from various bank APIs, and the async interface of `convert` results in a very roundabout way of exchanging from one currency to another in our codebase. A feature such as this would help us a lot (hence I've already implemented it), but also I've seen numerous discussions on why convert is asynchronous, so perhaps it would be interesting for others.

As part of this patch, I've prepared a `convertSync` method that expects the exchange rate endpoint to be a simple object. I'm opening this as a draft because I would like your input on whether you think this change is a good feature in upstream Dinero and how more polished I would have to make it to get it merged in case you approve.

In any case, thanks for taking the time to look at this.

## Compliance

- [x] My change isn't breaking (it doesn't cause existing functionality to change).
- [x] I have read the [CONTRIBUTING](https://github.com/dinerojs/dinero.js/blob/develop/CONTRIBUTING.md) document.
- [x] My code follows the coding style of this project.
- [x] I have properly formatted my commit messages with [cz-cli](https://github.com/commitizen/cz-cli), or manually, following the [Angular Commit Messages Guidelines](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines).
- [x] I have updated the documentation accordingly (or my changes doesn't require documentation changes).
- [x] I have added tests to cover my changes (or my changes doesn't require new tests).
- [x] I [added myself as a contributor](https://github.com/dinerojs/dinero.js/blob/develop/CONTRIBUTING.md#contributors).


sarahdayan:
Hi @vipera and thanks for your contribution. I'm sorry I didn't get to review it sooner.

With v2 out, I'm not looking to add new features to v1. I know you created this PR a while ago and I'm sorry it didn't get merged. However, I hope you'll still feel like contributing to v2 which foundation is much saner to add features.

nickpalmer:
This is called `exchange()` in @soldi/core.

There is a compatibility layer here for Dinero v1 users not interested in rewriting their software to Dinero v2:
https://www.npmjs.com/package/@soldi/dinero

PR #241, opened 2021-07-16
feat(dinero): add optional roundingMode parameter to Dinero.percentage

### LINKED ISSUE #208: Allow specifying roundingMode for Dinero.percentage
**Is your feature request related to a problem? Please describe.**

I was recently using Dinero to calculate percentages while needing to use a specific rounding mode (the context was that I was redoing calculations that had been previously done with a set rounding mode, so needed to faithfully reproduce the original calculations to be able to evaluate). I noticed that even though `percentage` uses `multiply` under the hood, unlike `multiply`, it does not support specifying a rounding mode to be used.

**Describe the solution you'd like**

`percentage` should have an optional `roundingMode` parameter.

**Describe alternatives you've considered**

One can also just use `multiply`, of course, but then what's the point of having the `percentage` convenience method?

**Additional context**

I noticed this idea was also brought up in a [comment](https://github.com/dinerojs/dinero.js/issues/144#issuecomment-653136024) in #144 when looking for duplicate issues before making this one.

Also, more than happy to make a PR implementing this. Just wanted to throw a feature request out first to make sure it's something the maintainers would want before submitting a PR!

seankwalker:
fix #208


sarahdayan (approved):
Looks good to me! Thanks for your contribution :)

PR #271, opened 2021-07-21
docs(add): mention "sum"

Haroenv:
this will give a relevant result for the "sum" search

### DIFF
--- a/website/data/docs/api/mutations/add.mdx
+++ b/website/data/docs/api/mutations/add.mdx
@@ -54,7 +54,7 @@ add(d1, d2); // a Dinero object with amount 144545 and scale 4
 
 ### Add more than two objects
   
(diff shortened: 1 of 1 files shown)


website/data/docs/api/mutations/add.mdx
```diff
@@ -53,6 +53,8 @@ add(d1, d2); // a Dinero object with amount 144545 and scale 4
 ```
 
 ### Add more than two objects
+  
+To retrieve the sum of multiple objects, you can call the "add" function multiple times, for example in a reduce:
```
sarahdayan, line 57:
```suggestion
To retrieve the sum of multiple objects, you can call the `add` function multiple times.
```

sarahdayan (approved):

PR #309, opened 2021-08-01
feat: provide better support for non-decimal currencies

### LINKED ISSUE #294: [Bug]: `toUnit` can't be accurate when used with big integers

### Current behavior

Function `toUnit` currently returns a `number`, whatever the amount type is.

```js
// ...
const d1 = dinero({ amount: 4545, currency: USD });
const d2 = dineroBigint({ amount: 4545n, currency: USD });
const d3 = dineroBigjs({ amount: new Big(4545), currency: USD });

toUnit(d1); // 45.45
toUnit(d2); // 45.45
toUnit(d3); // 45.45
```

To do so, Dinero.js uses the `toNumber` function exposed on the calculator to cast the generic amount into a number, then retrieve the unit format and return it.

The problem is that `number` isn't always fit to represent the unit format accurately.

For example, if the user has one of the following objects to represent amount $ 10000000000000000.50:

```js
// ...

const d = dineroBigint({ amount: 1000000000000000050n, currency: USD });
const d = dineroBigjs({ amount: new Big(1000000000000000050n), currency: USD });
```

Then `toUnit` can't format it correctly because this number can't be safely casted into a `number`.

A side-effect is that this inaccurate number is passed on to `toFormat`, which `transformer` function uses `toUnit` to expose the pre-formatted `amount`.

### Expected behavior

The function `toUnit` should return an accurate amount.

### Steps to reproduce

https://codesandbox.io/s/fervent-paper-flsvt?file=/main.js

### Version

v2.0.0-alpha.6

### Environment

All environments

### Code of Conduct

sarahdayan:
This PR changes the way Dinero.js does formatting, providing better support for non-decimal currencies and large integers. Many changes are going in this PR, as this touches several concepts of the library.

## Currencies now accept multi-bases

Currencies can have multiple bases. It's the case for the pre-decimal pound sterling, the _livre tournois_ of the French Old Regime, or the Harry Potter wizarding currency of Great Britain. Users creating custom currencies (e.g., for games) may also want to create such currencies.

Until now, Dinero.js allowed you to support such currencies by providing the number of units of the smallest subdivision to the biggest (e.g., 240 pence in a pound) as the `base`. It meant that [`toUnit` could only return a double representation of the amount](https://v2.dinerojs.com/docs/api/formatting/to-unit), and that [you'd have to write a lot of logic to properly format it](https://v2.dinerojs.com/docs/guides/formatting-non-decimal-currencies#handling-currencies-with-multiple-subdivisions).

Accepting multiple-bases lets users express this complexity and enables better formatting in the new `toUnits` function.

## `toUnit` is replaced with `toUnits`

The `toUnit` function currently returns a double, which is problematic at several levels:

- IEEE 754 doubles are unreliable, which is why Dinero.js avoids manipulating them altogether.
- It makes no sense to ever return a non-decimal currency in a decimal representation.
- Large integers can't be accurately casted to `number`s, so this operation can't work well with them.

The decimal representation (e.g., 10.5) is prevalent in systems using a decimal currency, but this is only a human representation, and only matters when you're displaying an amount on your site or app. Programming-wise, this representation isn't particularly helpful, and it's even limiting.

This PR removes `toUnit` and replaces it with `toUnits`, which returns each subdivision as an integer (a `number`, a `bigint`, a `Big`, [depending on what you use as the amount type](https://v2.dinerojs.com/docs/guides/using-different-amount-types)), in an array.

This accounts for decimal and non-decimal currencies, without forcing the "human" representation of one over the other. Thanks to multi-bases, the function handles the heavy lifting of distributing an amount into the right amount of each of its subdivision (e.g., 267 pence is 1 pound, 2 shillings and 3 pence, `toUnits` returns `[1, 2, 3]`). Since `toUnits` is now used by `toFormat` under the hood (instead of `toUnit`), it makes formatting even easier.

## `toFormat` is replaced with `toDecimal`

Until now, `toFormat` was the catch-all formatting function for any formatting need. It was also "only" a wrapper around `toUnit`, exposing the right information in a transformer function for users to format and display objects.

An intermediary design was to expose both the data from `toUnits` and a `decimal` representation in `toFormat`. But ultimately, there's no point in having a single function that exposes everything: it's computationally more expensive and it does too many things.

Instead, this PR introduces `toDecimal`, a formatter function that returns a stringified decimal representation. This is more reliable because of the string format which prevents the system from converting it to an inaccurate binary representation.

Another benefit is that the string representation retains the exponent, meaning that $10.50, which used to be returned as `10.5` will now be returned as `"10.50"`. If users need this as a double, despite the potential lack of accuracy (e.g., to manipulate it or use it with the Intl API), they can easily cast it with `Number` or `parseFloat`, but that's their responsibility.

`toDecimal` is only meant to be used with decimal, single-base objects. For multi-base or non-decimal needs, `toUnits` is recommended.

## `toUnits` and `toDecimal` take a transformer as its second parameter

```ts
declare function f<TAmount, TOutput>(d: Dinero<TAmount>, t: Transformer<TAmount, TOutput>): TOutput;
```

This makes it easier for users to customize the output based on the exposed `value`:
- A tuple in `toUnits`
- A stringified double in `toDecimal`

```js
const formatted = toDecimal(d, ({ value, currency }) => `${currency.code} ${value}`);

// Instead of

const decimal = toDecimal(d);
const formatted = `${toSnapshot(d).currency.code} ${decimal}`;
```

## PR notes

**BREAKING CHANGE:** the `toUnit` and the `toFormat` functions were removed.

fix #294


johnhooks:
@sarahdayan I read through the `toUnits` function and it seems like a much more useful function than the current function `toUnit`. What help do you need to push this forward?

johnhooks:
I've created a [branch](https://github.com/johnhooks/dinero.js/tree/dev) that merges this PR, the the api-extractor PR. I would like to use it, but I'm having a ton of trouble figuring out how to publish it in some way. To build my own version seems like it requires renaming all the packages and imports because everything is scoped to `@dinero.js` but for me to publish it would require being scoped to `@johnhooks`... I feel like I'm missing something.

packages/core/etc/core.api.md
```diff
@@ -51,11 +54,17 @@ export enum ComparisonOperator {
     // (undocumented)
     GT = 1,
     // (undocumented)
-    LT = -1
+    LT = -1,
 }
 
 // @public (undocumented)
-export function convert<TAmount>(calculator: Calculator<TAmount>): (dineroObject: Dinero<TAmount>, newCurrency: Currency<TAmount>, rates: Rates<TAmount>) => Dinero<TAmount>;
```
johnhooks, line 59: I've been noticing that this `api.md` files are being reformatted somewhere in the commit process. It looks like prettier should ignore `**/*.md` files but some how they are being reformatted. Is it reforming the `ts` code blocks inside the markdown? It would be better not to have them reformatted or reformatted before the commit. Right now it seems like the precommit hook formats documents but doesn't add them to the current commit. So the commit completes but the formatting changes are left over for the next commit or an amend.

sarahdayan: I've been poking around it for a few hours and for now can't find an explanation. Prettier does ignore these files so my guess is that the files might have been manually saved at some point, and some other local formatter impacted the files.

Another lead is line endings that may be different between environments. This can be solved with `.gitattributes`, I'll experiment with that.

sarahdayan:
@johnhooks I made a bunch of changes to the original PR and I think it's in a good state now. Would love your opinion and review on the current design.

johnhooks:
@sarahdayan I've started reading through it. So far, I like the concept of `toDecimal` and `toUnits`. Returning a string value from `toDecimal` makes sense to me.

johnhooks (approved):
@sarahdayan It looks super thorough. This is a big change and from what I can see you updated every file that was affected. I saw a few grammar choices I would make differently, all very small changes. I really appreciate your grasp of the code base. 

I didn't dive deep into the division operation source code or tests. I need to familiarize myself with them more.

I'm excited to use these changes.

packages/core/src/api/transformScale.ts
```diff
@@ -1,34 +1,31 @@
...
-import { greaterThan } from '../utils';
+import { down } from '../divide';
+import type { Calculator, Dinero, DivideOperation } from '../types';
+import { computeBase, greaterThan } from '../utils';
 
 export type TransformScaleParams<TAmount> = readonly [
   dineroObject: Dinero<TAmount>,
-  newScale: TAmount
+  newScale: TAmount,
+  divide?: DivideOperation
```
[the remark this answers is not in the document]
sarahdayan: I just realized that I forgot to export the provided divide functions and test them (they're unit tested in `core` but not as parameters in `dinero.js`) 🤦 Will fix this.

website/data/docs/getting-started/quick-start.mdx
```diff
@@ -127,17 +127,17 @@ const d4 = dinero({ amount: 1150, currency: USD });
 hasSubUnits(d4); // returns true
 ```
 
-You can display Dinero objects into any format, exactly the way you want. Dinero.js lets you build a formatter that exposes a pre-formatted amount, in major currency units.
+You can display Dinero objects into any format, exactly the way you want. Dinero.js lets you build a formatter that exposes a pre-formatted amount, in units and in stringified decimal format (if the currency is decimal and single-base).
```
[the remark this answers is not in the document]
sarahdayan: This was actually a leftover from the previous design with `toFormat`. I updated it in 762cd5e1.

website/data/docs/guides/integrating-with-payment-services.mdx
```diff
@@ -42,18 +42,18 @@ const response = await client.charges.create({
 
 ## Integrating with Paypal
 
-The [Paypal](https://www.paypal.com/) payment platform provides APIs to process payments and manage orders. Unlike most platforms, it expects a money representation with an amount in major currency units. You can use [`toUnit`](/docs/api/formatting/to-unit) to retrieve this value.
+The [Paypal](https://www.paypal.com/) payment platform provides APIs to process payments and manage orders. Unlike most platforms, it expects a money representation with an amount in major currency units. You can use [`toDecimal`](/docs/api/formatting/to-decimal) to retrieve this value.
```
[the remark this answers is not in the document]
sarahdayan: Hmm, "retrieve" works better here I believe because the user's goal is to "get" the value, not to format.

website/data/docs/guides/integrating-with-payment-services.mdx
```diff
@@ -42,18 +42,18 @@ const response = await client.charges.create({
...
-const { dinero, toSnapshot } = require('dinero.js');
+const { dinero, toSnapshot, toDecimal } = require('dinero.js');
 const { USD } = require('@dinero.js/currencies');
 
 function toPaypalMoney(dineroObject) {
   const { currency, scale } = toSnapshot(dineroObject);
 
   return {
-    value: toUnit(dineroObject).toFixed(scale),
+    value: toDecimal(dineroObject, ({ value }) => value),
```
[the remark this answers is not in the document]
sarahdayan: Good catch!

packages/dinero.js/src/api/__tests__/hasSubUnits.test.ts
```diff
@@ -13,84 +13,165 @@ import { hasSubUnits } from '..';
...
     });
   });
   describe('Big.js', () => {
     const dinero = createBigjsDinero;
     const bigjsUSD = castToBigjsCurrency(USD);
+    const bigjsMGA = castToBigjsCurrency(MGA);
+    const bigjsGBP = {
+      code: 'GBP',
+      base: [new Big(20), new Big(12)],
+      exponent: 1n,
```
johnhooks, line 121: @sarahdayan I found a typing issue.
```suggestion
      exponent: new Big(1),
```

website/data/docs/guides/integrating-with-payment-services.mdx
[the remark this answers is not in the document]
johnhooks: Sounds good, my logic was that `toDecimal` is one of the format functions.

johnhooks:
@sarahdayan I found something odd while porting this change to my Deno module, and it doesn't come up in this repo because test files aren't type checked.

https://github.com/dinerojs/dinero.js/blob/1061fb144d6669204f9ad0f6cfa4cc08838a2d32/packages/dinero.js/src/api/__tests__/transformScale.test.ts#L20

In all the tests, the `snapshot` variables returned from `toSnapshot` are of type `unknown`.

I think has to do with casting the return value of the default transform function, the type checker can't figure out what `TOuput` should be because it isn't being inferred.

https://github.com/dinerojs/dinero.js/blob/1061fb144d6669204f9ad0f6cfa4cc08838a2d32/packages/core/src/api/toSnapshot.ts#L8-L17

I'm playing around with it now trying to figure it out.

website/data/docs/guides/integrating-with-payment-services.mdx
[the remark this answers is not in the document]
sarahdayan: Maybe this could be rewritten into:

```suggestion
The [Paypal](https://www.paypal.com/) payment platform provides APIs to process payments and manage orders. Unlike most platforms, it expects a string representation with an amount in major currency units. You can use [`toDecimal`](/docs/api/formatting/to-decimal) to format the object and pass this value to Paypal.
```

johnhooks:
If I provide a custom transform function it resolves the type issue. It returns `DineroSnapshot<number>` rather than `unknown`.

```ts
const snapshot = toSnapshot(transformScale(d, 4), ({value}) => value);
```

johnhooks:
I have a grasp of the issue, but still don't know how to resolve it. Providing a default transformer creates a type issue, because a `TOutput` could be provided as a generic argument but not provide a transform function, then `TOutput` and the return type of the transformer conflict. 

```ts
const snapshot = toSnapshot<number, SomeWrongType>(d);
```

But when the casting the return of the default transformer argument, it's forcing the compiler to accept it. But when a transformer isn't provided there isn't any type information for the type system to collect because its been overridden by `as TOutput`.

I believe a condition generic would work, it needs to infer the return type of the transformer, or never be provided. Still working out the details.

johnhooks:
I still haven't found the answer, though it's making me question why `toDecimal`, `toSnapshot`, `toUnits` need a transformer, not because I'm having an issue with the types, but why do these functions need to return any arbitrary type?  It seems best for them to return specific types of `TAmount`, and the end user build what they need out of those building blocks, rather than complicate this functions.

```ts
function toDecimal<TAmount>(dineroObject: Dinero<TAmount>): string;

function toSnapshot<TAmount>(dineroObject: Dinero<TAmount>): DineroSnapshot<TAmount>;

function toUnits<TAmount>(dineroObject: Dinero<TAmount>): TAmount[];
```

Those are totally useful and specific functions.

website/data/docs/guides/integrating-with-payment-services.mdx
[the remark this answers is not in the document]
johnhooks: Very nice.

johnhooks:
@sarahdayan I put together my suggested changes in this commit, I don't believe I can push it to this remote branch. It only modifies the source code. If you approve these changes. I can work on updating all the documentation and examples.

https://github.com/dinerojs/dinero.js/compare/feat/to-units...johnhooks:dinero.js:feat/to-units

sarahdayan:
@johnhooks Fixed the type issue you found with overloads in [9662b64](https://github.com/dinerojs/dinero.js/pull/309/commits/9662b645025f2b979fc9fdf5454fc6b1801f2f0c) (with the kind help of @mattpocock), should be fine now.

johnhooks:
@sarahdayan that is perfect and I learned something new! It was really tripping my mind. It had to be possible, but I couldn't find good information about how a generic can be optional.

I'm still not sure it's entirely necessary for `toSnapshot` or `toUnits`, makes the most since for `toDecimal`. My preference would be to return simpler outputs, though since the types work now, it doesn't really matter lol. They still return what I expect and I am able to compose them the way I want.

sarahdayan:
>I'm still not sure it's entirely necessary for `toSnapshot` or `toUnits`, makes the most sense for `toDecimal`.

I dislike the idea of having inconsistent APIs for three functions that serve a similar purpose (formatting, whatever your need is). At the same time I'm not fond of introducing more surface API than necessary.

I guess it makes sense to remove it for `toSnapshot` given how low-level of a function it is, however this is more blurry for `toUnits` since this is meant to be used mostly to format non-decimal/multi-base currencies. It's basically the equivalent of `toDecimal` for thise currencies, so it doesn't sound like a good idea to treat it differently.

```js
const currencyLabelsMap = { MGA: ['ariary', 'iraimbilanja'] };

// With `transformer`
const formatted = toUnits(d, ({ value, currency }) => {
  const labels = currencyLabelsMap[currency.code];

  return value.map((amount, index) => `${amount} ${labels[index]}`).join(', ');
});

// Without `transformer`
const units = toUnits(d);
const { currency } = toSnapshot(d);
const labels = currencyLabelsMap[currency.code];
const formatted = units
  .map((amount, index) => `${amount} ${labels[index]}`)
  .join(', ');
```

Thoughts?

johnhooks:
I was thinking of `toUnits` as less of a format function and more as a utility to split an amount into currency divisions. It's used internally by `toDecimal`. But I understand your point, it's currently the only "format" function for non decimal currencies, but that isn't apparent.

packages/core/src/api/transformScale.ts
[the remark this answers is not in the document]
sarahdayan: Fixed in [97fcbe1](https://github.com/dinerojs/dinero.js/pull/309/commits/97fcbe1e0c633a203d6f20b9aa6e29f210346093) and [56efcc2](https://github.com/dinerojs/dinero.js/pull/309/commits/56efcc2a37f81f789dc80927038c6923709d3031).

sarahdayan:
@johnhooks I removed the `transformer` fro `toSnapshot` and left it for `toUnits` and `toDecimal`.

All tests for `transformScale` are added.

We should be good now, wdyt?

johnhooks:
@sarahdayan is this still the title and url you want for this link?

https://github.com/dinerojs/dinero.js/blob/80fdc0b2ac7e6971672012552ebf9d2d7f7290a4/website/data/docs/core-concepts/formatting.mdx?plain=1#L81-L82

johnhooks:
@sarahdayan, another `to-format` reference. I know there are redirects, but do you still want to use the old url? Also `toUnits` doesn't have a `decimal` property?

https://github.com/dinerojs/dinero.js/blob/80fdc0b2ac7e6971672012552ebf9d2d7f7290a4/website/data/docs/guides/formatting-non-decimal-currencies.mdx?plain=1#L39-L43

website/data/docs/api/formatting/to-decimal.mdx
```diff
@@ -0,0 +1,61 @@
...
+
+<Parameters>
+
+<Parameter name="dineroObject" type="Dinero<TAmount>" required={true}>
+
+The Dinero object to format.
+
+</Parameter>
+
+<Parameter name="transformer" type="Transformer<TAmount, TOutput>" required={false}>
```
johnhooks, line 27:
```suggestion
<Parameter name="transformer" type="Transformer<TAmount, TOutput, TValue>" required={false}>
```

website/data/docs/api/formatting/to-decimal.mdx
```diff
@@ -0,0 +1,61 @@
...
+const d1 = dinero({ amount: 1050, currency: USD });
+const d2 = dinero({ amount: 10545, currency: USD, scale: 3 });
+
+toDecimal(d1); // "10.50"
+toDecimal(d2); // "10.545"
+```
+
+### Use a custom transformer
+
+If you need to further transform the snapshot before returning it, you can pass a custom function.
```
johnhooks, line 52:
```suggestion
If you need to further transform the value before returning it, you can pass a custom function.
```

website/data/docs/api/formatting/to-units.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+<Parameters>
+
+<Parameter name="dineroObject" type="Dinero<TAmount>" required={true}>
+
+The Dinero object to format.
+
+</Parameter>
+
+<Parameter name="transformer" type="Transformer<TAmount>" required={false}>
```
johnhooks, line 23:
```suggestion
<Parameter name="transformer" type="Transformer<TAmount, TOutput>" required={false}>
```

website/data/docs/api/formatting/to-units.mdx
```diff
@@ -0,0 +1,86 @@
+---
+title: toUnits
+description: Get the amount of a Dinero object in units.
+returns: TAmount[]
```
johnhooks, line 4:
```suggestion
returns: TOutput = TAmount[]
```

website/data/docs/api/formatting/to-units.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+const GBP = { code: 'GBP', base: [20, 12], exponent: 1 };
+const d = dinero({ amount: 267, currency: GBP });
+
+toUnits(d); // [1, 2, 3]
+```
+
+### Use a custom transformer
+
+If you need to further transform the snapshot before returning it, you can pass a custom function.
```
johnhooks, line 70:
```suggestion
If you need to further transform the value before returning it, you can pass a custom function.
```

I think the work "value" works better here, otherwise there could be confused that the return value should be simular to `toSnapshot` which has a very specific (and different) return type.

johnhooks (changes requested):
Looks great, other than the few touch ups to the docs to clear out old references and tighten up a few of the parameters and return types in the api documentation.

sarahdayan:
@johnhooks I fixed the missed `toFormat` references, thanks for catching them!

packages/core/src/api/toUnits.ts
```diff
@@ -0,0 +1,43 @@
...
+  transformer?: Transformer<TAmount, TOutput, readonly TAmount[]>
+];
+
+export function toUnits<TAmount, TOutput>(calculator: Calculator<TAmount>) {
+  const getDivisorsFn = getDivisors(calculator);
+
+  return function toUnitsFn(
+    ...[
+      dineroObject,
+      transformer = ({ value }) => value as TOutput,
```
johnhooks, line 15: I don't like the `as` in there. I've been thinking about these functions a lot and the return of the default transformer isn't supposed to be `TOutput` with how the declaration written.  

If a `transformer` isn't provided then return the correct value that is intended and not use a default argument and infer `TOutput`.

So just return the `TAmount[]` in `toUnits` and in `string` in `toDecimal`. It doesn't have to go through a default `transformer`. Leaving the `as` in there just opens up the possibility of new issues in the future during refactors.

```suggestion
      transformer,
```

packages/core/src/api/toUnits.ts
```diff
@@ -0,0 +1,43 @@
...
+    ]: ToUnitsParams<TAmount, TOutput>
+  ) {
+    const { amount, currency, scale } = dineroObject.toJSON();
+    const { power, integerDivide, modulo } = calculator;
+
+    const bases = isArray(currency.base) ? currency.base : [currency.base];
+    const divisors = getDivisorsFn(bases.map((base) => power(base, scale)));
+
```
johnhooks, line 23:
```suggestion

    if (!transformer) return divisors;
    
```

packages/core/src/api/toDecimal.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
+  const toUnitsFn = toUnits<TAmount, readonly TAmount[]>(calculator);
+  const computeBaseFn = computeBase(calculator);
+  const equalFn = equal(calculator);
+
+  return function toDecimalFn(
+    ...[
+      dineroObject,
+      transformer = ({ value }) => value as TOutput,
```
johnhooks, line 21:
```suggestion
      transformer,
```

packages/core/src/api/toDecimal.ts
```diff
@@ -0,0 +1,61 @@
...
+    const isDecimal = !isMultiBase && isBaseTen;
+
+    // eslint-disable-next-line functional/no-expression-statement
+    assert(isDecimal, NON_DECIMAL_CURRENCY_MESSAGE);
+
+    const units = toUnitsFn(dineroObject);
+
+    const getDecimalFn = getDecimal(dineroObject.formatter);
+    const value = getDecimalFn(units, scale);
+
```
johnhooks, line 41:
```suggestion

    if (!transformer) return value;
```

johnhooks (changes requested):
I don't like the `as` in `toUnits` or `toDecimal`. I've been thinking a lot about these functions, and the return of the default transformer isn't supposed to be `TOutput` with how the declaration written.  

If a `transformer` isn't provided then return the correct value that is intended and not use a default `transformer` argument because `TOutput` can't be inferred in that way.

So just return the `TAmount[]` in `toUnits` and `string` in `toDecimal`. It doesn't have to go through a default `transformer`. Leaving the `as` in there just opens up the possibility of new issues in the future during refactors.

Aside: Is there anyway I could make suggestions in a commit? Writing suggestions like this in the browser seems super error prone.

This is a branch with my refactor [7b14803](https://github.com/johnhooks/dinero.js/commit/7b14803bae1bc4c71bc4beacf467935f2d4925ab)

packages/core/src/api/toDecimal.ts
```diff
@@ -0,0 +1,61 @@
...
+export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
+  const toUnitsFn = toUnits<TAmount, readonly TAmount[]>(calculator);
+  const computeBaseFn = computeBase(calculator);
+  const equalFn = equal(calculator);
+
+  return function toDecimalFn(
+    ...[
+      dineroObject,
+      transformer = ({ value }) => value as TOutput,
+    ]: ToDecimalParams<TAmount, TOutput>
```
johnhooks, line 22:
```suggestion
    ...[dineroObject, transformer]: ToDecimalParams<TAmount, TOutput>
```

packages/core/src/api/toUnits.ts
```diff
@@ -0,0 +1,43 @@
...
+];
+
+export function toUnits<TAmount, TOutput>(calculator: Calculator<TAmount>) {
+  const getDivisorsFn = getDivisors(calculator);
+
+  return function toUnitsFn(
+    ...[
+      dineroObject,
+      transformer = ({ value }) => value as TOutput,
+    ]: ToUnitsParams<TAmount, TOutput>
```
johnhooks, line 16:
```suggestion
    ...[dineroObject, transformer]: ToUnitsParams<TAmount, TOutput>
```

sarahdayan:
@johnhooks That works for me, and makes the build slightly smaller by not setting default values. Check https://github.com/dinerojs/dinero.js/pull/309/commits/2bf6376ddbac503865217c9fbcc4069e2da12274#diff-df69e9b5f173d135b3648f576277a76610d36de0bcb071773625b9ddc3889a3b and let me know what you think :)

johnhooks:
Looks great, I'd say it's ready!

PR #448, opened 2021-10-19
fix(trimScale): check for zero value in countTrailingZeros

reckter:
`trimSccale`, when called with a `0` as the `amount`, kept in a endless loop.

Cause:
`countTrailingZeros` was stuck, as there never was any non-zero value to brake the loop.

I am unsure wether this is the cleanest solution, but  it at least fixes the problem.

Also added a quick test for regression.


packages/core/src/utils/countTrailingZeros.ts
```diff
@@ -11,6 +11,9 @@ export function countTrailingZeros<TAmount>(
   const equalFn = equal(calculator);
 
   return (input: TAmount, base: TAmount) => {
+    if (equalFn(calculator.zero(), input)) {
+      return calculator.zero();
+    }
     const zero = calculator.zero();
```
sarahdayan, line 17:
```suggestion
    const zero = calculator.zero();

    if (equalFn(zero, input)) {
      return zero;
    }
```

packages/core/src/api/__tests__/trimScale.ts
```diff
@@ -0,0 +1,20 @@
+import { calculator } from '@dinero.js/calculator-number';
```
sarahdayan, line 1: There's already a [test suite](https://github.com/dinerojs/dinero.js/blob/main/packages/dinero.js/src/api/__tests__/trimScale.test.ts) for `trimScale`, could you move the tests there?

packages/core/src/api/__tests__/trimScale.ts
```diff
@@ -0,0 +1,20 @@
...
+  it('should not crash on 0 amounts', () => {
+    const result = trim(dinero({ amount: 0, currency: EUR }));
+    expect(toSnapshot(result)).toEqual({ amount: 0, scale: 2, currency: EUR });
+  });
+  it('should calculate scale correctly', () => {
+    const result = trim(
+      dinero({ amount: 100000000, scale: 10, currency: EUR })
+    );
+    expect(toSnapshot(result)).toEqual({ amount: 1, scale: 2, currency: EUR });
+  });
```
sarahdayan, line 19: Since `trimScale` is already tested (see previous comment), this one might not be necessary.

packages/core/src/utils/countTrailingZeros.ts
```diff
@@ -11,6 +11,9 @@ export function countTrailingZeros<TAmount>(
   const equalFn = equal(calculator);
 
   return (input: TAmount, base: TAmount) => {
+    if (equalFn(calculator.zero(), input)) {
```
sarahdayan, line 14: Could you add a test for this branch in [this test suite](https://github.com/dinerojs/dinero.js/blob/main/packages/core/src/utils/__tests__/countTrailingZeros.test.ts)?

sarahdayan (changes requested):
Thanks @reckter, good catch! Let me know if you need further help.

packages/core/src/api/__tests__/trimScale.ts
reckter: I did not find that at all, thanks! will do :)

reckter:
@sarahdayan I left the comments "unresolved", as I wasn't sure how you handle these on PR's here. I've should've fixed all of them, so feel free to resolve them :)

reckter:
Hey, is there anything I can do to help get this merged? :)

sarahdayan (approved):
We're good to go! Thanks @reckter I'll deploy this soon.

svarlet:
@sarahdayan or any other maintainer, can you please trigger the release pipeline?

johnhooks:
@svarlet and @reckter I don't have access to release this. This solution may not be ideal, but until its released, I created a fork and published a modified version that includes this PR and the [feat: provide better support for non-decimal currencies](https://github.com/dinerojs/dinero.js/pull/309). So the API is different than 2.0.0-alpha.8, and I haven't attempted to deploy the updated docs, so you would need to look at the docs in the `website/data/docs` folder. I went through the hassle because I really wanted the that pull request, it adds rounding options to `transformScale`. But this PR was also included.

the repo is https://github.com/johnhooks/dinero.js

the package on npm is `@bitmachina/dinero.js`

PR #582, opened 2022-02-08
Adding all Currencies to also have Currency<bigint>

reckter:
Adding an additional `168-bigint` to the `amendments` folder. that contain all currencies in `Currency<bigint>` format.
I am not sure wether this is the right place or even way of doing it, but as they are currently missing, I thought providing one proposal on how to solve this might be a good idea.

Let me know your thoughts on this, happy to move the folder / or do something completely different :)


johnhooks (commented):
I agree, adding BigInt currencies makes sense. Though for packaging as an umd, separating the `number` and `BigInt` currencies seems like a better idea to me, just like the calculators.

sarahdayan:
Yes, creating two packages in the same fashion as the calculators is probably the way to go.

johnhooks:
I just created [PR #665](https://github.com/dinerojs/dinero.js/pull/665) it moves the currencies packages to currencies-number. If @sarahdayan thinks this is the way forward, @reckter could you modify this PR to create a new currencies-bigint package?

It avoid duplication, would making a script to generate the currency files be a better approach? All the data could be in a single json file.

sarahdayan:
@johnhooks Totally agree that we should generate the currencies. The `currency` package could actually be repurposed into it, holding a single TypeScript or JSON file with the definitions. Users could reuse it to generate currency packages for custom amount data structure.

reckter:
@johnhooks Looking at #670, it should be fairly trivial with that to implement this PR, so will look into that after it has been merged. (Feel free to just go ahead yourself, If I take to long 😅 ). Thanks for picking this up again 🍺 !

johnhooks:
Sounds good!

sarahdayan:
Thanks for this PR, I picked up the project again and included bigint currencies in https://github.com/dinerojs/dinero.js/commit/0e4aef8524ac3ec0d294a0b90fbbef329801cbe1. I marked you both as co-authors, thanks for your work!

PR #628, opened 2022-04-26
fix: prevent string input to be used in allocate function

### LINKED ISSUE #588: [Bug]: allocate() does not throw when given numeric strings as input

### Current behavior

First off, thanks for the great library.

.allocate() function has unexpected behaviour when given falsy input.

### Expected behavior

I expected an error:
`` TypeError: You must provide a non-empty array of numeric values greater than 0 ``

### Steps to reproduce

```
try {
    Dinero({ amount: 100 })
    .allocate([ '2', '98' ])
} catch (error) {
    console.log(error)
}
```

It never finishes nor throws meanwhile hogging memory.

The same for this:
``.allocate([ 1, '98' ])``

But this one yields the expected TypeError:
``.allocate([ 'a', '98' ])``

### Version

1.9.1

### Environment

14.17.3

### Code of Conduct

matheusfcorocher:
## How was it solved?
It was fixed by adding a condition that verifies if ratio is numeric and has number type in areValidRatios function. 

fixes #588

### DIFF
--- a/src/services/helpers.js
+++ b/src/services/helpers.js
@@ -33,6 +33,7 @@ export function isPercentage(percentage) {
 export function areValidRatios(ratios) {
   return (
     ratios.length > 0 &&
+    ratios.every(ratio => typeof ratio === 'number' && isNumeric(ratio)) &&
     ratios.every(ratio => ratio >= 0) &&
     ratios.some(ratio => ratio > 0)
   )
--- a/src/services/helpers.js
+++ b/src/services/helpers.js
@@ -33,8 +33,7 @@ export function isPercentage(percentage) {
 export function areValidRatios(ratios) {
   return (
     ratios.length > 0 &&
-    ratios.every(ratio => typeof ratio === 'number' && isNumeric(ratio)) &&
-    ratios.every(ratio => ratio >= 0) &&
+    ratios.every(ratio => typeof ratio === 'number' && isNumeric(ratio) && ratio >= 0) &&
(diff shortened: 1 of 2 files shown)


src/services/helpers.js
```diff
@@ -33,6 +33,7 @@ export function isPercentage(percentage) {
 export function areValidRatios(ratios) {
   return (
     ratios.length > 0 &&
+    ratios.every(ratio => typeof ratio === 'number' && isNumeric(ratio)) &&
     ratios.every(ratio => ratio >= 0) &&
```
sarahdayan, line 37: We can combine these two to avoid traversing the whole array twice.

test/unit/dinero.spec.js
```diff
@@ -277,6 +277,10 @@ describe('Dinero', () => {
       const shares = Dinero({ amount: 1003 }).allocate([0, 50, 50])
       expect(shares.map(share => share.getAmount())).toEqual([0, 502, 501])
     })
+    test('should allocate the amount of the Dinero object into new ones and distribute the remainder when one of ratios is float number', () => {
+      const shares = Dinero({ amount: 1000 }).allocate([1.618, 1])
+      expect(shares.map(share => share.getAmount())).toEqual([619, 381])
+    })
```
sarahdayan, line 283: This seems unrelated to the PR's objective.

sarahdayan (approved):

sarahdayan:
Thanks @matheusfcorocher!

PR #642, opened 2022-07-08
docs: fixes example of calculating with scale

thchia:
Thank you for this library.

I believe there is an error in the docs in the example of calculating objects of different scales. I have validated this by running the code shown in the example and do indeed get a different result than what is shown there.


johnhooks (approved):
I agree, that is correct

sarahdayan (approved):

PR #646, opened 2022-07-26
feat(#645) enhance toFormat to support generic return types

### LINKED ISSUE #645: Feature Request: Enhance toFormat to support generic return types
# Use Case
We ran into a scenario where we want to format a `Dinero` object in a user-specified locale but not to a single `string`, but to its component parts - e.g. localized currency symbol, localized numeric amount in the localized order (e.g. currency symbol may prefix or suffix the numeric amount).

The reason for doing that is so we can use a library such as [react-native-currency-input](https://www.npmjs.com/package/react-native-currency-input) where we can specify separate `prefix` and `suffix` components to render the localized currency symbol separate from the monetary numeric value.

Currently the [toFormat()](https://v2.dinerojs.com/docs/api/formatting/to-format) function only supports a `Transformer` function which returns a `string`, the `Transformer` can't return a generic return type, even though there does not appear to be any reason not to.

# Existed Related Support
A common way of localizing monetary values is to use [Intl.NumberFormat#format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format) which returns a single `string` and so can already be used with `toFormat`:

```ts
const d = dinero({ amount: 500, currency: USD });

const formattedResult = toFormat(d, ({ amount, currency }) => {
  const formatter = new Intl.NumberFormat('fr-CA', {
    style: 'currency',
    currency: currency.code,
  });
  
  return formatter.format(amount);
});

// formattedResult = "5,00 $ US"
```

However `Intl.NumberFormat` also provides a [formatToParts()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) function which localizes and formats a monetary value and returns an array of component parts that can be rendered individually as desired.

# Desired Solution

What is desired is for `toFormat()` to support a generic return type so that the following is also supported:

```ts
const d = dinero({ amount: 500, currency: USD });

const formattedResult = toFormat(d, ({ amount, currency }) => {
  const formatter = new Intl.NumberFormat('fr-CA', {
    style: 'currency',
    currency: currency.code,
  });
  
  return formatter.formatToParts(amount);
});

// formattedResult = [{"type":"integer","value":"5"}, {"type":"decimal","value":","}, {"type":"fraction","value":"00"}, {"type":"literal","value":" "}, {"type":"currency","value":"$ US"}] 
```

For backwards compatibility that generic return type can default to a string, but the passed in formatter could return a different result object type if desired, like the above does.

# Contribution

The changes required to support this use case are minimal and are fully backwards compatible (as far as I can tell) and so I've implemented it myself in #646 and submitted for review.

It's worth pointing out that this PR purely contribues type changes, and doesn't introduce any runtime changes.

I would love to see this merged in as this was motivated by a near-term upcoming user story to implement a localized `TextInput` field where the user just edits the monetary amount and the currency symbol is localized appropriately and rendered separately. There are also later stories that we believe will also make use of this.

danrivett:
Closes #645.

### DIFF
--- a/packages/dinero.js/src/api/__tests__/toFormat.test.ts
+++ b/packages/dinero.js/src/api/__tests__/toFormat.test.ts
@@ -2,6 +2,7 @@ import { USD } from '@dinero.js/currencies';
 import Big from 'big.js';
 
 import { toFormat } from '..';
+import type { Transformer } from '../../..';
 import {
   castToBigintCurrency,
   castToBigjsCurrency,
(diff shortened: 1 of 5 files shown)


johnhooks (approved):
This seems like a reasonable change to me. I reviewed the source, new tests and documentation. It all looks good.

website/data/docs/api/formatting/to-format.mdx
```diff
@@ -41,6 +41,31 @@ const d = dinero({ amount: 500, currency: USD });
 toFormat(d, ({ amount, currency }) => `${currency.code} ${amount}`); // "USD 5"
 ```
 
+### Format to a complex result object
+
+Although formatting to a single string value is the most common scenario, there are other times (such as when more complex rendering is required) where it is desirable to format to a more complex return type.
```
sarahdayan, line 46:
```suggestion
Although formatting to a string is the most common scenario, there are times when you need to format to a different type.
```

website/data/docs/api/formatting/to-format.mdx
```diff
@@ -41,6 +41,31 @@ const d = dinero({ amount: 500, currency: USD });
...
+
+For example [Intl.NumberFormat#formatToParts(number)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) returns each formatted component in an ordered array to allow for more complex rendering options.
```
sarahdayan, line 48:
```suggestion
For example [Intl.NumberFormat#formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts) returns the formatted string as chunks in an array to allow for more complex manipulations.
```

website/data/docs/api/formatting/to-format.mdx
```diff
@@ -41,6 +41,31 @@ const d = dinero({ amount: 500, currency: USD });
...
+
+This could be used to display the currency symbol separately from the numeric value in an input control, as follows:
```
sarahdayan, line 50:
```suggestion
For example, you could use this to display the currency symbol and code as a complex input element.
```

website/data/docs/api/formatting/to-format.mdx
```diff
@@ -41,6 +41,31 @@ const d = dinero({ amount: 500, currency: USD });
...
+// More complex formatting - returning separate formatted parts
+toFormat(d, ({ amount, currency }) => {
+  const formatter = new Intl.NumberFormat('fr-CA', {
+    style: 'currency',
+    currency: currency.code,
+  });
+
+  return formatter.formatToParts(amount);
+}); // [{"type":"integer","value":"5"}, {"type":"decimal","value":","}, {"type":"fraction","value":"00"}, {"type":"literal","value":" "}, {"type":"currency","value":"$ US"}]
+```
```
sarahdayan, line 67: This example could be even more compelling with a visual example (e.g., React code) that displays something like this:

Capture d’écran 2022-11-07 à 23 11 53

What do you think? If this is too time consuming, I can contribute on your PR.

packages/dinero.js/src/api/__tests__/toFormat.test.ts
```diff
@@ -11,58 +12,164 @@ import {
 } from '../../../../../test/utils';
 
 describe('toFormat', () => {
+  const simpleFormatter = ({ amount, currency }) =>
```
sarahdayan, line 15: This is more explicit IMO.

```suggestion
  const stringFormatter = ({ amount, currency }) =>
```

packages/dinero.js/src/api/__tests__/toFormat.test.ts
```diff
@@ -11,58 +12,164 @@ import {
...
 
-      expect(toFormat(d, formatter)).toBe('USD 5');
+        expect(toFormat(d, simpleFormatter)).toBe('USD 4.545');
+      });
     });
-    it('formats the Dinero object with the passed transformer using the scale', () => {
-      const formatter = ({ amount, currency }) => `${currency.code} ${amount}`;
-      const d = dinero({ amount: 4545, currency: USD, scale: 3 });
+    describe('with complex formatter', () => {
+      const complexFormatter: Transformer<
```
sarahdayan, line 34:
```suggestion
      const arrayFormatter: Transformer<
```

sarahdayan, line 34: You don't need to type the formatter, type inference is enough here. This should allow you to hoist this instead of recreating it in the `bigint` and Big.js tests.

danrivett:
Thanks for the reviews @johnhooks and @sarahdayan. I will implement these actions in the next week, I'm just in the middle of a story so want to get that wrapped up first before context switching.

sarahdayan:
Hey @danrivett, thanks for your contribution! Unfortunately I'll have to close it—`toFormat` was removed in [`v2.0.0-alpha.11`](https://github.com/dinerojs/dinero.js/releases/tag/v2.0.0-alpha.11) in favor of [`toDecimal`](https://v2.dinerojs.com/docs/api/formatting/to-decimal) which supports a generic output.

I hope you contribute again soon.

PR #665, opened 2022-11-06
feat: move @dinero.js/currencies to @dinero.js/currencies-number

johnhooks:
This PR facilitates creating a new package @dinero.js/currencies-bigint and should help [PR #582](https://github.com/dinerojs/dinero.js/pull/582) move forward.


johnhooks:
@sarahdayan it looks like this breaks the CircleCI script

sarahdayan:
> @sarahdayan it looks like this breaks the CircleCI script when trying to move dist folders to the workspace.

Yup, that's because you need to update `.circleci/config.yml` which still refers to the old `currencies` directory into `currencies-number`.

I noticed there are other leftover references to update in:
- `bundlesize.config.json`
- `CONTRIBUTING.md`
- `packages/currencies-number/README.md`
- `packages/dinero.js/package.json`
- `website/data/docs/core-concepts/currency.mdx`
- `website/data/docs/faq/does-dinerojs-support-cryptocurrencies.mdx`
- `website/data/docs/getting-started/quick-start.mdx`
- `website/data/docs/getting-started/upgrade-guide.mdx`
- `website/utils/console/init.ts`

Feel free to also rebase onto `main` to pass the Semantic PR check.

johnhooks:
> Feel free to also rebase onto `main` to pass the Semantic PR check.

How can I change the commit titles? I noticed they weren't right when the check didn't pass.

sarahdayan:
> How can I change the commit titles?

You don't have to, I [changed the rules](https://github.com/dinerojs/dinero.js/commit/99e95b317df8a61bc441933ae4607e2ede12116a) so that only the PR title must be semantic. When we squash and merge, GitHub will pick the PR title for the squashed commit. All you need to do is rebase your branch onto the latest verison of `main`.

packages/currencies-number/scripts/genCurrencies.mjs
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable no-console, functional/no-expression-statement */
```
johnhooks, line 1: This file is hacky, I made it a script rather than part of a package because it doesn't seem like it fits into the normal purpose of the library, its just a helper to reduce duplication of the currency code. I did check in the generated code, because I think it's important for it to be viewable to the end users.

packages/currencies-number/src/iso4217/amendments/168/aed.ts
```diff
@@ -2,6 +2,8 @@ import type { Currency } from '@dinero.js/core';
 
 /**
  * United Arab Emirates dirham.
+ *
+ * @public
```
johnhooks, line 6: I added @public to all these for api-extractor

scripts/getCurrencies.mjs
```diff
@@ -0,0 +1,24 @@
+/**
```
johnhooks, line 1: I thought about putting this in `@dinero.js/currencies` but then it would need to be built and need all the overhead of a package when its just a script to get the JSON file. Though if there is a way to import a json file directly from a package, I would be open to that and remove this file.

data/iso4217/amendments/168.json
```diff
@@ -0,0 +1,628 @@
+[
```
johnhooks, line 1: I put this file here because I couldn't figure out how to import a json directly from a package, so I just put it in the top level.

packages/currencies-number/src/iso4217/amendments/168/aed.ts
public: Thanks for tagging me, this library looks great actually :D

johnhooks: lol sorry I'll write `@public` next time

public: No problem, I love discovering new stuff like this!

sarahdayan: @public We love new contributors around here!

sarahdayan:
This PR is getting really big, we'd benefit from splitting it into smaller ones.

Here's a suggestion of how to break it down:

1. Move `@dinero.js/currency` into `@dinero.js/currency-number`.
2. Create a script in `scripts/` to generate the currency files using Node.js APIs. At first, this can just generate `number` versions.
3. Update the `prepare` commands to run the generation script.
4. Delete the existing currency files and ignore them from version control.
5. Modify the script to generate `bigint` currencies in a new `@dinero.js/currency-bigint` package.

What do you think?

johnhooks:
I can do all of that.

sarahdayan:
@johnhooks Also, I started thinking more deeply about the split of packages and while generating currencies fully make sense, we may not need to have two separate packages per currency type if we provide multiple endpoints using the `browser` entry of `package.json`.

Let's maybe do what's necessary to generate the currencies, but table the split into two packages for now and instead explore this simpler solution. Basically skipping step 1, and leave step 5 to #582.

Sorry for the change of heart, new ideas came AFK. Before we commit, what are your thoughts?

johnhooks:
That's perfectly okay. I can change direction. Most of my time on this was figuring out how to walk an estree and use `@babel/traverse` to grab all the currencies descriptions from the comments of the original source. That was a super hack throw away script, and now it's all collected in the `168.json` file.

johnhooks:
@sarahdayan whats the best way to walk back practically all the changes, but not lose them just in case making separate packages ends up being the way to go? Is opening up a PR for `feat: generate currencies` be an okay?

johnhooks:
If I branch off this and [change the branch the PR points to](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request), will all the history on GitHub be lost?

sarahdayan:
So, for extra clarity, the new process we're going for is:

1. Create a script in `scripts/` to generate the currency files using Node.js APIs.
2. Update the `prepare` commands to run the generation script. Normally, running the command shouldn't change anything because we'd be re-generating existing files identically.
3. Delete the existing currency files and ignore them from version control.
4. Add the `@public` note in the currency generator.

Each thing can be its own PR.

This gives us more time to think of package strategy, yet still we make progress on currency generation.

Sounds good @johnhooks?

johnhooks:
That sounds good. I agree, this is a better route.

PR #666, opened 2022-11-06
build: add TypeScript project references

johnhooks:
This is just an idea, it makes the TypeScript compiler a bit smarter. It won't rebuild packages that don't need it. Also it should allow the TypeScript language server to discover package types in development without requiring a rebuild of the types.

This change also allows doing some neat stuff like importing '@dinero.js/test' for the test utils, rather than a long path to get to the top directory in each test.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -14,6 +14,7 @@ export default {
     __TEST__: true,
   },
   moduleNameMapper: {
+    '^@dinero.js/test$': '<rootDir>/test/utils/',
     '^@dinero.js/(.*)$': '<rootDir>/packages/$1/src/',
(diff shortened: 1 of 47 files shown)


jest.config.ts
```diff
@@ -14,6 +14,7 @@ export default {
     __TEST__: true,
   },
   moduleNameMapper: {
+    '^@dinero.js/test$': '<rootDir>/test/utils/',
```
sarahdayan, line 17: I'm afraid this could be confusing for contributors who discover the codebase and don't check this file, because it has exactly the same scheme as legitimate dependencies.

We could keep it simple and just map this path to `'test-utils'`, wdyt?

sarahdayan (approved):
LGTM!

PR #667, opened 2022-11-06
build: add turbo repo

### LINKED ISSUE #671: [Bug]: build steps are run twice during CI

### Current behavior

During local development I've notice when running `yarn install` it will call the `prepare` scripts of all the workspace packages. Is this intended behavior? 

This also happens in the CI build pipeline. So `babel`, `tsc` and `api-extractor` are all run twice. Because they are run after `yarn install` and also during `yarn build`. What is the right way to remove this duplication?

>Certain script names are special. If defined, the preinstall script is called by yarn before your package is installed. For compatibility reasons, scripts called install, postinstall, prepublish, and prepare will all be called after your package has finished installing.

[^^ reference](https://classic.yarnpkg.com/en/docs/package-json#toc-scripts)

### Expected behavior

Build steps should not be duplicated in CI or local development.

### Steps to reproduce

`yarn install`
`yarn build`

### Version

2.0.0-alpha.9

### Environment

Node.js 16.13.0

### Code of Conduct

johnhooks:
TurboRepo caches all the build, lint, test runs and wont rerun if the inputs haven't changed. This makes rebuilding a single package much faster. There will definitely be more to consider.

Fixes #671

### Use TurboRepo over Lerna to manage the monorepo

In order for TurboRepo to properly cache all the build steps, each package needs to implement their own scripts, otherwise a change in a single package will invalidate the cache for the whole build step.

The current build process runs `eslint` and `jest` from the root directory.

### Building

- Split `build:tsc` and `build:types`, no longer call `yarn run build:tsc` from the `build:types` script so they can be separate cached processes.
- Remove the individual `build` scripts from the packages, it is now managed by TurboRepo.

### Linting

- To lint from inside each package directory, `eslint` needs to find the config and ignore files in the root, which requires cli flags, kinda of ugly but works.

### Testing

- To test from inside each package directory, `jest` requires a config inside the project.
- To simplify the requirement above, a new `jest.base.ts` was added to split all the shared configuration settings.
- An important new setting in the root `jest.config.ts` is the projects array. It allows `jest` to still be run at a top level and run each project individually. But the script `test` doesn't run `jest` this way, it calls it from each package folder.
  ```ts
  projects: [
    '<rootDir>/packages/calculator-bigint/jest.config.ts',
    '<rootDir>/packages/calculator-number/jest.config.ts',
    '<rootDir>/packages/core/jest.config.ts',
    '<rootDir>/packages/currencies/jest.config.ts',
    '<rootDir>/packages/dinero.js/jest.config.ts',
  ],
  ```

### `package.json`

- Modified the workspaces property, it was required for TurboRepo to find the packages of the monorepo. I'm not sure what that changes for the examples or website.

### `lint-staged.config.js`

- Call `eslint` directly, otherwise the filenames are passed down to each package, the relative paths aren't correct and have many files included that didn't change for that specific package

### `ship.config.js`

- Switch to using `package.json` as the main version file, because of no longer requiring the `lerna.json` file.
- **I need help on what would need to change in the `versionUpdated` lifecycle function.** Is `yarn lerna version ...` how all the packages versions are updated?

### Add `turbo.json`

- This file describes the build process. It uses dependencies, inputs, and outputs to determine whether or not there is a cache miss because of a change.


johnhooks:
I brought [PR #666](https://github.com/dinerojs/dinero.js/pull/666) into this PR. Otherwise the build system couldn't run concurrency because it couldn't find the types if they hadn't been built.

website/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint",
-    "postinstall": "yarn --cwd ../ install && yarn --cwd ../ build"
```
johnhooks, line 10: I believe running the full build after install is causing an issue in CircleCI

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ aliases:
       - ~/.cache/yarn
   - &run_yarn_install
     name: Install dependencies
-    command: yarn install --frozen-lockfile
+    command: yarn install --frozen-lockfile --ignore-scripts
```
johnhooks, line 20: This was the problem. Without --ignore-scripts, yarn would run the prepare script, which would trigger a whole build and timeout the install process.

johnhooks:
Also I believe if the `.turbo` folders are cache during ci it will dramatically speed up the the actions.

sarahdayan:
I don't know why you decided to close this, but I'd be very inclined to migrate to Turborepo eventually, all the experience I've had with it so far is wonderful.

It's still quite some processes to migrate, so it's okay to do it in more iterative chunks.

johnhooks:
I'll focus on just building all the outputs of the dist folder, that's the step that takes the longest. With jest and eslint I had to do some weird stuff so they could be run as a script in the package folders.

**Issues**

- jest and eslint scripts for the individual packages. jest requires a config file in each package, eslint just needs cli arguments to find the config and ignore files.

- CircleCI workspaces aren't the right way to store the cached files because they are only passed on to upstream jobs.

- CircleCI caching could work, but I need to figure out how to generate a cache key.

So far, I haven't been able to get the cache to work and for Turborepo to get a cache hit in ci, it always reruns.

sarahdayan:
@johnhooks Maybe let's try to fix the performance issues you found _with_ the existing setup, and we can benchmark whether Turborepo really helps later on.

johnhooks:
@sarahdayan, issue #671 could be resolved by adding Turborepo, then it wouldn't matter that a build step is repeated in multiple jobs. Which solution would you prefer spending time on?

sarahdayan:
@johnhooks Is it because of the way Turborepo does caching? Would it work without a Vercel remote caching subscription though, if the builds are run in different jobs in a CI?

johnhooks:
@sarahdayan Turborepo seems designed to work best with a Vercel remote caching subscription. It would work with this GitHub action [turborepo-gh-artifacts](https://github.com/felixmosh/turborepo-gh-artifacts). It creates a local cache server that saves the cache as GitHub artifacts. Figuring out how to reimplement it in CircleCI is probably more time that I can commit to the issue. Saving the whole folder of Turborepo files to a CircleCI cache would work, if I can figure out how to name the key so it will use the same cache if the files haven't changed. Turborepo docs state that it creates a hash of all the effected files, but I think in a monorepo it creates separate ones for every task of every package. So, it's kinda complicated.

If there is something like CircleCI workspaces that allow the current job to reattach a folder and read the previously saved contents, that would work. But workspaces only propagate to upstream jobs... which would help when rerunning tasks during a single CI workflow, just like the issue we are having with the `prepare` script and `yarn install`.

johnhooks:
@sarahdayan I think this one is ready for review.

It actually resolves the `prepare` issue. `prepare` is removed from all the packages, through there is a top level `prepare`. They had to be removed from the individual packages to avoid rerunning step.

Does having a top level `prepare` resolve the issue with the documentation and website?

johnhooks:
@sarahdayan I updated the description to explain the changes.

johnhooks:
@sarahdayan I pulled to get the update you made to this branch, and it created a merge... What would have been the right thing to do, rebase?

.circleci/config.yml
```diff
@@ -18,22 +18,40 @@ aliases:
   - &run_yarn_install
     name: Install dependencies
     command: yarn install --frozen-lockfile
-  - &restore_dist_folders
-    name: Restore dist folders
+  - &restore_build_folders
+    name: Restore build folders
```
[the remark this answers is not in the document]
johnhooks: Because it's no longer just the `dist` folders, it's a bunch of different build artifacts. But I can definitely rename it.

sarahdayan:
> I pulled to get the update you made to this branch, and it created a merge... What would have been the right thing to do, rebase?

@johnhooks No worries, a merge is fine because we'll end up squashing the PR into a single commit anyway at the end.

turbo.json
```diff
@@ -0,0 +1,61 @@
...
+    "build": {
+      "dependsOn": ["build:esm", "build:umd:cjs", "build:types"],
+      "outputs": ["dist/**"]
+    },
+    "lint": {
+      "inputs": ["**/*.ts", "**/*.js"],
+      "outputs": []
+    },
+    "prepare": {
+      "dependsOn": ["^prepare"],
```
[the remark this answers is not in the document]
sarahdayan: I'm asking because I don't see any `prepare` command in packages, there's just one in the root `package.json`.

jest.base.ts
```diff
@@ -0,0 +1,27 @@
...
+  // It is overridden by the toplevel jest config
+  moduleNameMapper: {
+    '^test-utils$': '<rootDir>/../../test/utils/',
+    '^@dinero.js/(.*)$': '<rootDir>/../$1/src/',
+    '^dinero.js$': '<rootDir>/../dinero.js/src/',
+  },
+  transform: {
+    // '\\.ts': 'babel-jest',
+    '\\.ts': ['babel-jest', { rootMode: 'upward' }],
+  },
```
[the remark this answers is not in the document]
sarahdayan: Gotcha. Can we remove the comment in l. 24?

turbo.json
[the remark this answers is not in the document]
johnhooks: Oh, sorry, it must have been in preparation for #670, it would require running a prepare script and order could matter. I can remove it and add it back if necessary.

ship.config.js
```diff
@@ -15,16 +14,6 @@ module.exports = {
   publishCommand({ tag }) {
     return `yarn publish --access public --tag ${tag}`;
   },
```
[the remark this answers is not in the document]
johnhooks: Was it previously done by `lerna`?

turbo.json
```diff
@@ -0,0 +1,61 @@
+{
+  "$schema": "https://turborepo.org/schema.json",
+  "pipeline": {
+    "build:clean": {
+      "inputs": ["lib/**/*", "dist/**/*"],
+      "outputs": []
+    },
+    "build:tsc": {
```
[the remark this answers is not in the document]
johnhooks: I'll have to change it in all the `packages/*` too, that okay?

website/package.json
johnhooks: I'll try adding it as part of the workspace and we'll see how it works.

.circleci/config.yml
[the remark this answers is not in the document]
sarahdayan: No that makes sense!

website/package.json
johnhooks: I had trouble adding the website to the workspace. Though I added this command back in, and everything is working as it did previously. I think the initial issue was caused by not filtering the build process to just the `packages` folder, with `turbo build --filter=./packages/*`, so it was running build in the website in CIi. Though that shouldn't happen now with the proper filtering.

package.json
```diff
@@ -1,35 +1,34 @@
...
+    "packages/calculator-bigint",
+    "packages/calculator-number",
+    "packages/core",
+    "packages/currencies",
+    "packages/dinero.js",
+    "examples/cart-react",
+    "examples/cart-vue",
+    "examples/pricing-react",
+    "examples/starter"
+  ],
```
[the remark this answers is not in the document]
johnhooks: I also simplified it to:
https://github.com/dinerojs/dinero.js/blob/dfc9fc04b98178dee11e119f1d98cc9d29320eab/package.json#L6-L9

johnhooks:
@sarahdayan I did a rebase to merge most the commits, except the recommendations. Rebasing onto `upstream/main` was getting to be a real pain with 44 commits that applied changes and then reverted them.

sarahdayan:
> The last script I haven't moved from `lerna` to `turbo` is `watch`. How do you use that script, so I can make sure the match the functionality in `turbo`?

@johnhooks This is an obsolete script that only made sense when I didn't had added `moduleNameMapper` to `jest.config.js` and thus was using build files in tests instead of sources.

So `watch` and `on:change` scripts can go away, and we can update the `CONTRIBUTING.md` guide to [update the note about `watch`](https://github.com/dinerojs/dinero.js/blob/c5b91730b6db00128554f6d03d3a82be12390899/CONTRIBUTING.md#tests).

sarahdayan:
> The last script I haven't moved from `lerna` to `turbo` is `watch`. How do you use that script, so I can make sure the match the functionality in `turbo`?

@johnhooks This is an obsolete script that only made sense when I didn't had added `moduleNameMapper` to `jest.config.js` and thus was using build files in tests instead of sources.

So `watch` and `on:change` scripts can go away, as well as `start` in the root `package.json`, and we can update the `CONTRIBUTING.md` guide to [update the note about `watch`](https://github.com/dinerojs/dinero.js/blob/c5b91730b6db00128554f6d03d3a82be12390899/CONTRIBUTING.md#tests).

johnhooks:
@sarahdayan I can't see why the Vercel build failed, or rerun it.

sarahdayan:
Nice! You have a few conflicts to solve because of the #698 merge but it should be easy.

johnhooks:
@sarahdayan the last thing to consider is the release pipeline. From what I can tell, the output of `shipjs prepare --dry-run` looks good. But I don't know how to test actually releasing it.

scripts/build-types.mjs
```diff
@@ -0,0 +1,28 @@
...
+  'build:types',
+  '--filter=./packages/*',
+  ...args,
+  ...apiExtractorArgs,
+])
+  .then(() => process.exit(0))
+  .catch((err) => {
+    console.error(err);
+    process.exit(1);
+  });
```
johnhooks, line 28: This is similar to how remix does their build scripts. It allows for more fine tuned control.

scripts/exec.mjs
```diff
@@ -0,0 +1,37 @@
...
+    child.on('error', handleError);
+    child.on('close', (code) => {
+      if (code === 0) {
+        resolve(void 0);
+      } else {
+        reject(new Error(`${command} exited with code ${code}`));
+      }
+    });
+  });
+}
```
johnhooks, line 37: Only things I did differently than remix was to inherit the io from the parent process, and also exit the child if the parent does. Otherwise if you try to cancel the build process with `^C` the child will keep running.

lint-staged.config.js
```diff
@@ -2,7 +2,7 @@
 module.exports = {
   '**/*.ts': () => 'yarn test:types',
   '!(website/)**/*.(js|mjs|ts|)': (filenames) =>
-    `yarn lint --fix ${filenames.join(' ')}`,
+    `eslint --ext .js,.ts --fix ${filenames.join(' ')}`,
```
johnhooks, line 5: This is necessary because otherwise `turbo` runs `eslint` in each project not from the root.

package.json
```diff
@@ -47,6 +47,7 @@
     "babel-plugin-module-resolver": "4.1.0",
     "big.js": "6.2.1",
     "bundlesize": "0.18.1",
+    "cross-spawn": "7.0.3",
```
johnhooks, line 50: This seems to be pretty popular, some other dependency must already require the exact same version because it didn't modify the lock file at all when this was added.

I think it'll help if someone is developing on windows and tries to build the project.

packages/calculator-bigint/etc/calculator-bigint.api.md
```diff
@@ -14,9 +14,7 @@ export const add: BinaryOperation<bigint>;
 // @public (undocumented)
 export const calculator: {
     add: BinaryOperation<bigint, bigint>;
-    compare: BinaryOperation<bigint,
-    ComparisonOperator
-    >;
+    compare: BinaryOperation<bigint, ComparisonOperator>;
```
johnhooks, line 17: Fixes the reformatting that happened in [e7e9a19](https://github.com/dinerojs/dinero.js/commit/e7e9a19e6eb8e4ff8903867a60c1457a8d241d0c). We'll need to figure out what is going on. I didn't notice the reformatting when I reviewed the PR for this commit. But there were a lot of changes. It should be easier to notice in small PRs, and I will keep a look out for it.

scripts/build-types.mjs
```diff
@@ -0,0 +1,40 @@
...
+//   if (args.includes('--local')) {
+//     const index = args.indexOf('--local');
+//     apiExtractorArgs = [
+//       ...apiExtractorArgs.slice(0, index),
+//       ...apiExtractorArgs.slice(index + 1, apiExtractorArgs.length),
+//     ];
+//   }
+// } else if (!args.includes('--local')) {
+//   apiExtractorArgs = [...apiExtractorArgs, '--local'];
+// }
```
johnhooks, line 33: Once we are ready for `api-extractor` to enforce the api in the `api.md` files, we just need to uncomment this and it will apply `--local` only if the `CI` env variable isn't truthy.

packages/calculator-bigint/package.json
```diff
@@ -36,7 +36,7 @@
     "build:clean": "rimraf ./{dist,temp}",
     "build:esm": "babel src --root-mode upward --extensions '.ts' --out-dir dist/esm --ignore '**/*/__tests__/'",
     "build:definitions": "tsc -b",
-    "build:types": "api-extractor run --local",
+    "build:types": "node ../../scripts/build-types.mjs",
```
johnhooks, line 39: All the packages call the build-types script rather than `api-extractor` directly.

scripts/build-types.mjs
```diff
@@ -0,0 +1,29 @@
...
+  url.fileURLToPath(new URL('.', import.meta.url)), // __dirname es module style
+  '../node_modules/.bin/api-extractor'
+);
+
+// Strip `--local` out. It will be configured only by this script.
+const args = process.argv.slice(2).filter((arg) => arg !== '--local');
+
+// When in CI, `api-extractor` shouldn't apply the local flag, this script will error if an api
+// change has occurred without being committed in an `api.md` file.
+// const localFlag = process.env.CI ? [] : ['--local'];
```
johnhooks, line 17: Once we are ready for `api-extractor` to enforce the api we just need to uncomment the `localFlag`.

johnhooks:
@sarahdayan the website builds locally and I can't see the error details on Vercel. Can you look into what's happening?

sarahdayan:
> @sarahdayan the website builds locally and I can't see the error details on Vercel. Can you look into what's happening?

```txt
Error: The file "/vercel/path0/website/.next/routes-manifest.json" couldn't be found. This is normally caused by a misconfiguration in your project.
Please check the following, and reach out to support if you cannot resolve the problem:
1. If present, be sure your `build` script in "package.json" calls `next build`.
2. Navigate to your project's settings in the Vercel dashboard, and verify that the "Build Command" is not overridden, or that it calls `next build`.
3. Navigate to your project's settings in the Vercel dashboard, and verify that the "Output Directory" is not overridden. Note that `next export` does **not** require you change this setting, even if you customize the `next export` output directory.
Learn More: https://err.sh/vercel/vercel/now-next-routes-manifest
```

Did the output directory change somehow?

johnhooks:
@sarahdayan what are the build settings on Vercel? Specifically the build command and root directory?

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-console, no-process-exit, functional/no-expression-statement */
```
sarahdayan, line 1: We can disable these rules in [`.eslintrc.js`](https://github.com/johnhooks/dinero.js/blob/6b1abb2854aa1845ed9ef61c392c0feb256d9816/.eslintrc.js) using the `overrides` option with `files: ['scripts/**/*']`.

scripts/utils/exec.mjs
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-void, no-console, functional/no-expression-statement, jsdoc/require-description, valid-jsdoc */
+import { spawn } from 'cross-spawn';
+
+/**
+ * Async exec.
+ *
+ * Inspired by `remix-run` build scripts, [reference](https://github.com/remix-run/remix/blob/534e1ec071f17654a4db8622e30d6ff70548ce26/scripts/build.mjs).
```
sarahdayan, line 7:
```suggestion
 * Async `exec`
 *
 * Inspired by `remix-run` build scripts
 * https://github.com/remix-run/remix/blob/534e1ec071f17654a4db8622e30d6ff70548ce26/scripts/build.mjs
```

scripts/utils/exec.mjs
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param {string} command - Shell command to execute.
+ * @param {string[]} [args] - Command arguments.
+ * @param {import('node:child_process').SpawnOptions} options - Options to pass to `cross-spawn`.
+ * @returns {Promise} Returns prom.
```
sarahdayan, line 12:
```suggestion
```

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
...
+import path from 'path';
+import * as url from 'url';
+
+import { exec } from './utils/exec.mjs';
+
+const apiExtractorPath = path.resolve(
+  url.fileURLToPath(new URL('.', import.meta.url)), // __dirname es module style
```
sarahdayan, line 8: Since we use this in 3 scripts we could have a `dirname` function in `utils` to abstract it.

```js
function dirname(path) {
  return url.fileURLToPath(new URL(path, import.meta.url));
}
```

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
...
+  '../node_modules/.bin/api-extractor'
+);
+
+// Strip `--local` out. It will be configured only by this script.
+const args = process.argv.slice(2).filter((arg) => arg !== '--local');
+
+// When in CI, `api-extractor` shouldn't apply the local flag, this script will error if an api
+// change has occurred without being committed in an `api.md` file.
```
sarahdayan, line 16:
```suggestion
// When in CI, `api-extractor` shouldn't apply the local flag. This script will
// error if an API change occurs without being committed in an `api.md` file.
// @todo Uncomment once we're ready to enforce the API
```

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
...
+// const localFlag = process.env.CI ? [] : ['--local'];
+
+/**
+ * ApiExtractor cli args.
```
sarahdayan, line 20:
```suggestion
 * ApiExtractor CLI arguments
```

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
...
+ */
+const apiExtractorArgs = ['run', '--local', ...args /* , ...localFlag */];
+
+console.log(`api-extractor ${apiExtractorArgs.join(' ')}`);
+
+exec(apiExtractorPath, [...apiExtractorArgs])
```
sarahdayan, line 26: Why do we need to spread `apiExtractorArgs` into a new array? Does `exec` mutate them?

scripts/lint.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+const eslintPath = path.resolve(
+  url.fileURLToPath(new URL('.', import.meta.url)),
+  '../node_modules/.bin/eslint'
+);
+
+const args = process.argv.slice(2);
+
+/**
+ * ESLint cli args.
```
sarahdayan, line 15:
```suggestion
 * ESLint CLI arguments
```

scripts/lint.mjs
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * ESlint needs to be run in each project for `turbo` to cache the results. Though it doesn't look
+ * up the file tree for the .eslintrc or .eslintignore files.
```
sarahdayan, line 18:
```suggestion
 * ESlint needs to be run in each project for `turbo` to cache the results, so
 * it doesn't look up the file tree for `.eslintrc` or `.eslintignore`.
```

scripts/test.mjs
```diff
@@ -0,0 +1,31 @@
...
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url));
+
+// Jest needs to be run from the monorepo root directory.
+const rootDir = path.resolve(__dirname, '../');
+
+const jestPath = path.resolve(__dirname, '../node_modules/.bin/jest');
+
+const args = process.argv.slice(2);
+
+// When in CI, `jest` should only use 4 workers.
```
sarahdayan, line 16:
```suggestion
// When in CI, Jest should only use 4 workers to avoid out of memory errors.
```

scripts/test.mjs
```diff
@@ -0,0 +1,31 @@
...
+const maxWorkers = process.env.CI ? ['--maxWorkers=4'] : [];
+
+/**
+ * Jest cli args.
```
sarahdayan, line 20:
```suggestion
 * Jest CLI arguments
```

ship.config.js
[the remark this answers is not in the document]
sarahdayan: Yes, here's what it does ([source](https://github.com/lerna/lerna/blob/main/commands/version/README.md)):

> 1. Identifies packages that have been updated since the previous tagged release.
> 2. Prompts for a new version.
> 3. Modifies package metadata to reflect new release, running appropriate [lifecycle scripts](https://github.com/lerna/lerna/blob/main/commands/version/README.md#lifecycle-scripts) in root and per-package.
> 4. Commits those changes and tags the commit.
> 5. Pushes to the git remote.

This should be handled by Ship.js, but we'll see how it goes in the next release.

sarahdayan (commented):
Thank you so much for your hard work on this one @johnhooks. This looks fantastic.

I left a bunch of comments but this is looking good. I'm going to look at what's going on in Vercel.

sarahdayan:
@johnhooks I think I found what's going on with Vercel.

Because this is now a Turborepo project, Vercel runs `cd .. && npx turbo run build --filter={./website}...` instead of `yarn run build` to build the project. But because `website` isn't a package the `build` script isn't called, so the `.next` directory doesn't exist.

johnhooks:
@sarahdayan it was all building before I added the `scripts/{test,lint,build-types}` scripts. I'm going to roll back and reconsider those. I might just use regular shell scripts, calling `exec` and figuring out the current working directory seems like a complexity that might not be worth it.

sarahdayan:
@johnhooks You don't need to rollback, we can make it work simply by making `website` a package. I'm on it.

sarahdayan:
Something odd is going on with `yarn install` that I'm not able to reproduce locally for now. I'll sleep it over and look at it tomorrow with fresh eyes.

scripts/build-types.mjs
johnhooks: If that is in the `scripts/utils` directory won't that always resolve to that directory, not the one it's called from?

scripts/build-types.mjs
johnhooks: I think originally I had it like this:

```js
exec(apiExtractorPath, [...apiExtractorArgs, ...args, ...localFlag])
```

But then I did a quick edit to disable the local Flag and moved everything up into the `apiExtractorArgs` so I could print them (otherwise you couldn't tell what was going on in the logs). So no it does not need to be spread into `exec`.

scripts/build-types.mjs
sarahdayan: Ah yes :facepalm: We can keep it that way, I'll suggest local edits for consistency.

.eslintrc.js
```diff
@@ -49,6 +49,21 @@ module.exports = {
...
     },
+    {
+      files: ['scripts/**/*'],
+      rules: {
+        'no-void': 'off',
+        'no-console': 'off',
+        'no-process-exit': 'off',
+        'functional/no-expression-statement': 'off',
+        '@typescript-eslint/naming-convention': 'off',
+        'import/no-extraneous-dependencies': 'off',
```
sarahdayan, line 60: Why do we need this rule?

scripts/build-types.mjs
```diff
@@ -0,0 +1,31 @@
+import path from 'path';
+import * as url from 'url';
+
+import { exec } from './utils/exec.mjs';
+
+const apiExtractorPath = path.resolve(
+  url.fileURLToPath(new URL('.', import.meta.url)), // __dirname es module style
```
sarahdayan, line 7:
```suggestion
  url.fileURLToPath(new URL('.', import.meta.url)),
```

scripts/build-types.mjs
sarahdayan: @johnhooks This comment might have been missed.

scripts/build-types.mjs
johnhooks: I totally agree it's ugly and not self explanatory of what is happening, that is why I had a comment.

.eslintrc.js
johnhooks: It's not necessary, I copy and pasted the override for `test/**/*` and it was in there, I forgot to remove it.

.eslintrc.js
```diff
@@ -49,6 +49,21 @@ module.exports = {
...
+        'jsdoc/require-description': 'off',
+        'jsdoc/match-description': 'off',
+        'jsdoc/require-description-complete-sentence': 'off',
```
johnhooks, line 63: Have you noticed `eslint-plugin-jsdoc` is really buggy in VSCode and stales the process sometimes?

johnhooks:
@sarahdayan by adding `website` to the `workspaces` should the `yarn.lock` in the `website` directory be deleted, since its being managed at the root level?

johnhooks:
Also if you change the Vercel build command to:

```sh
cd .. && npx turbo build '--filter=./packages/*' --filter=./website
```

Can we stop running  `yarn build` in the `prepare` script? If we can eliminate that, it would ease some of the work I'm doing to speed up CI.

scripts/build-types.mjs
johnhooks: Got it now.

johnhooks:
@sarahdayan I'm sorry I broke Vercel again and cant see whats wrong.

.eslintrc.js
sarahdayan: Does it? I never saw this happen. Shall we fix it separately?

sarahdayan:
> @sarahdayan I'm sorry I broke Vercel again and cant see whats wrong.

I reproduce inconsistently on my machine, but also when going back to 2230e907f31ddcfd46d4b6b3de34d525f8843f9c. From what I can see there seems to be caching issues where Turborepo ends up not building the UMD package for Dinero.js, causing the files that the website needs not to be there, and the build fails. I don't know how this happens, and can't reproduce it deterministically.

As you can see, my latest commit passed on Vercel 🤷‍♀️ 

> 
>     * Should there still be `website/yarn.lock`? Because adding `website` to the workspaces has caused yarn to add everything for the website into the root level `yarn.lock`

Good call, the dependencies were actually hoisted when I made the website a workspace. I removed `website/yarn.lock` in be0cadcb27c25d047b07d01cfbb6b8120e26944d.

> 
>     * The install command on Vercel would need to be changed to `cd .. && yarn install`.

Vercel does it for you. When you specify the root directory (here, `website`) you can include source files outside of it, and Vercel automatically `cd`s back to the actual project's root and installs dependencies.

johnhooks:
@sarahdayan I figured out way it was not consistently. The website didn't have `dinero.js` as a dependency. So turbo didn't know to build it after `dinero.js` was complete.

johnhooks (commented):
@sarahdayan I went through everything a final time. I made a few minor edits, notes are in the commit message.

ship.config.js
```diff
@@ -1,9 +1,14 @@
...
-    mainVersionFile: 'lerna.json',
-    // We rely on Lerna to bump our dependencies.
-    packagesToBump: [],
+    mainVersionFile: 'package.json',
+    packagesToBump: [
+      'packages/calculator-bigint',
+      'packages/calculator-number',
+      'packages/core',
+      'packages/currencies',
+      'packages/dinero.js',
```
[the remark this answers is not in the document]
sarahdayan: Yes, makes sense, thanks for adding it!

.eslintrc.js
sarahdayan: I don't use it.

sarahdayan (approved):

sarahdayan:
@johnhooks I took one last look and it looks good to me. I updated `ship.config.js` to use globs in [f46cd22](https://github.com/dinerojs/dinero.js/pull/667/commits/f46cd22d736cc91e61c6bf6d3ab0dce4f4e6296c) and weirdly, the site build failed (although the actual build completed successfully, but the manifest wasn't found). I ran it again to see if it was consistent, and both locally and in Vercel, it passed 🤷‍♀️ 

Not sure what happened but this seems to be a cache issue again. In the failed deploy, Turborepo hit the cache while in my redeploy, it bypassed it.

Site deployment is independent from bundle deployment so I'm fine merging and fixing the site issues separately. **How does it sound?**

johnhooks:
@sarahdayan I'm okay with that. I'll keep searching for the issue.

PR #670, opened 2022-11-08
refactor: generate currency source files

johnhooks:
As a first step to add currencies of different `TAmounts`, I created scripts to generate source files from a shared json data file. This should make implementing [PR #582](https://github.com/dinerojs/dinero.js/pull/582) trivial.

**Other changes**

- removed all original `Currency<number>` source files
- added `generate` script to `@dinero.js/currencies`
- updated `prepare` script of `@dinero.js/currencies` to generate sources files before building
- added `@public` comment to all `Currency<number>` objects for api-extractor
- added `@public` comment to `Currency<TAmount>` type for api-extractor


johnhooks:
Since api-extractor was added before this PR, we can be petty confident that this didn't make any changes to the existing api, because it didn't modify the `currencies.api.md` file generated by api-extractor.

johnhooks:
To verify the generation makes no changes to existing files, I replaced all the original source files from the main branch. After
running the generation script the replaced currency files are identical to the originals.

I modified the `genCurrenciesNumber.mjs` script by remove the `@public` comment for api-extractor. This was all that was required to make the generated code identical to the originals.

sarahdayan (commented):
Really cool change, that's gonna be useful.

One early thought before I review this more thoroughly—way the script is designed, if we add a `genCurrenciesNumber` script, we'll read and parse the JSON file twice (since we'll call `genCurrencies` in each script). We can avoid that by creating configuration objects instead that we pass to a generator. The generator reads and parses once, and maps over configurations to which it passes the data. That's how Rollup works, and it makes the final script easy to read and to update.

For example, we could have the entry point look like this:

```js
const numberCurrencyConfig = createCurrencyConfig(
  ({ currency, ...description }) => ({
    ...currency,
    description,
    genericType: 'number',
  })
);

const bigintCurrencyConfig = createCurrencyConfig(
  ({ currency, ...description }) => ({
    ...currency,
    description,
    base: `${base}n`,
    exponent: `${exponent}n`,
    genericType: 'bigint',
  })
);

const configs = [numberCurrencyConfig, bigintCurrencyConfig];

export function build() {
  // load data

  configs.forEach((config) => {
    // do stuff
  });
}
```

johnhooks:
Beautiful. I'll incorporate that.

scripts/genCurrenciesNumber.mjs
```diff
@@ -0,0 +1,30 @@
...
+  ),
+  // Source format function
+  ({ base, code, description, exponent }) =>
+    `
+      import type { Currency } from '../../../types';
+
+      /**
+       * ${description}.
+       */
+      export const ${code}: Currency<number> = {
```
reckter, line 24: This could easily be adopted to be even more generic (abstractions all the way!), By letting the `number` part (and the formatting of the numbers down below), be configured.

This is just a pop-up thought, and might just be a bit too much abstraction layering though.

scripts/genCurrencies.mjs
```diff
@@ -0,0 +1,76 @@
...
+    .catch(() => false);
+
+/**
+ * Generate a set of currency source files.
+ *
+ * @param {string} outputDir - The directory to output the source files into.
+ * @param {formatFn} formatFn - The format function used to generate individual currency files.
+ * @returns {Promise<void>} - Returns a Promise that resolves when all the files are generated.
+ */
+export async function genCurrencies(outputDir, formatFn) {
```
reckter, line 36: Because it confused me:
`format` might not be the ideal name here.
I got confused, as  to why there are to functions named `format` in this function being used.
One being prettier, the other one being a generator of code.

I would rename this to `generatefFn` instead, to be extra clear

scripts/genCurrencies.mjs
```diff
@@ -0,0 +1,76 @@
...
+    );
+
+  // Generate index ts file
+  const indexModuleText = currencies
+    .sort((a, b) => a.code.localeCompare(b))
+    .reduce(
+      (acc, current) =>
+        `${acc}export * from './${current.code.toLowerCase()}';\n`,
+      ''
+    );
```
reckter, line 68: Veeeeery nitpicky, I just don't like reduce calls all that much

```suggestion
    .map((current) => `export * from './${current.code.toLowerCase()}';`)
    .join("\n");
```

scripts/genCurrencies.mjs
johnhooks: `generateFn` sounds like a much better name, I will switch to that.

scripts/genCurrencies.d.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable functional/no-mixed-type */
```
johnhooks, line 1: I got tired of JSDoc types

scripts/genCurrencies.mjs
```diff
@@ -20,54 +22,103 @@ import { getCurrencyData } from './getCurrencyData.mjs';
...
+}
+
+const numberCurrencyConfig = createCurrencyConfig({
+  genericType: 'number',
+  mapNumber: (n) => `${n}`,
+});
+
+const bigintCurrencyConfig = createCurrencyConfig({
+  genericType: 'bigint',
+  mapNumber: (n) => `${n}n`,
```
johnhooks, line 61: @reckter I implemented it, I was just doing it to test, but I pushed it with everything else. Hopefully thats okay!

scripts/genCurrencies.mjs
```diff
@@ -20,54 +22,103 @@ import { getCurrencyData } from './getCurrencyData.mjs';
...
   const prettierOptions = {
     ...(await resolveConfig(process.cwd())),
     parser: 'babel',
   };
 
+  const configs = [numberCurrencyConfig, bigintCurrencyConfig];
+
+  // Generate index ts files
+  const indexPromises = configs.map(async ({ genericType }) => {
+    const outputDir = path.join(baseOutputDir, genericType);
```
johnhooks, line 81: I'm using the name of the `genericType` to create the folder structure. I don't know if this is best, but it works for these two types.

scripts/genCurrencies.mjs
```diff
@@ -20,54 +22,103 @@ import { getCurrencyData } from './getCurrencyData.mjs';
 // eslint-disable-next-line import/no-named-as-default-member
 const { format, resolveConfig } = prettier;
 
+const baseOutputDir = path.resolve(
+  url.fileURLToPath(new URL('.', import.meta.url)), // __dirname es module style
+  '../packages/currencies/src/iso4217/amendments/168'
```
johnhooks, line 27: I think eventually we would even want this broken out into another layer of abstraction, so it will auto generate other amendments. But I think leaving that as another step is okay.

johnhooks:
Oh and now the codesandbox doesn't build because of prepare, because the generate function wasn't called the source code isn't there. So maybe removing the old prepare as suggested in #671, and adding a new `"prepare": "generate"` would be a good idea.

johnhooks:
I'm unsure why the `ci/codesandbox` failed. Are all the packages built concurrently? There is a type error for all the currency files imported in the index file, stating they are not modules. I don't receive the same error locally.

packages/currencies/src/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './iso4217/amendments/168';
+export * from './iso4217/amendments/168/number';
```
johnhooks, line 1: I made the default to load the number currency library.

scripts/genCurrencies.mjs
```diff
@@ -0,0 +1,139 @@
...
+  await Promise.all(currencyPromises);
+}
+
+const start = performance.now();
+
+build()
+  // eslint-disable-next-line  promise/always-return
+  .then(() => {
+    const elapsed = Math.floor((performance.now() - start) / 10) / 100;
+    const msg = `Dinero.js currency generation successful\n😸  Done in ${elapsed}s`;
```
johnhooks, line 133: Added this to emulate api-extractors output. It's helpful to see in the CI output. Also added cat face, if too cute I can remove it.

scripts/genCurrenciesNumber.mjs
johnhooks: @reckter checkout out the abstraction now. Creating the currency generation config adds a `mapNumber` function which converts the JS number type of the json data file into a string that will be output into the generated source files.

Example for `bigint`

```ts
createCurrencyConfig({
  genericType: 'bigint',
  mapNumber: (n) => `${n}n`,
}
```

scripts/genCurrencies.mjs
reckter: That's very ok! :D

johnhooks:
Merging #672 fixed the issue I was having with `ci/codesandbox`, I don't know exactly how, but there must have been a step that was happening out of the correct order when the `prepare` scripts also had the redundant build steps.

sarahdayan:
Do you want to change the target branch to `johnhooks:remove-ci-duplication` to remove the unwanted diff? GitHub should automatically change it to main once #672 is merged, and this way we can review it without the noise.

johnhooks:
Sorry @sarahdayan its only letting me choose from the difference branches of the `dinerojs/dinero.js` repo not my fork. A little reading on Stack Overflow and it seems like maybe I need to open another pull request. Or maybe the easiest, I can walk back the merge?

sarahdayan:
@johnhooks Or let's just wait to merge #672 before we tackle this one.

johnhooks:
Sounds good, this is the more complicated change. I didn't think about merging it into this branch and how it would complicate your ability to make sense of the diff. I won't do that again.

johnhooks:
@sarahdayan I walked back the merge. There is still an issue to resolve in `ci/codesandbox` but at least you can view the file changes without the clutter of the other PR.

johnhooks:
@sarahdayan adding a build command to `./codesandbox/ci.sjon` fixed the issue. The build process must have been out of sync without running it.

johnhooks:
@sarahdayan this PR is ready for review.

data/iso4217/amendments/168.json
```diff
@@ -0,0 +1,628 @@
+[
+  {
+    "code": "AED",
+    "base": 10,
+    "exponent": 2,
+    "description": "United Arab Emirates dirham"
+  },
+  { "code": "AFN", "base": 10, "exponent": 2, "description": "Afghan afghani" },
+  { "code": "ALL", "base": 10, "exponent": 2, "description": "Albanian lek" },
+  { "code": "AMD", "base": 10, "exponent": 2, "description": "Armenian dram" },
```
johnhooks, line 10: @sarahdayan I would prefer them to all be formatted the same, would it be acceptable to ignore this file in prettier?

johnhooks:
@sarahdayan I'm still trying to figure out exactly why `ci/codesandbox` fails. I'm pretty sure it's running a build step out of order. `tsc` is being run in a package that depends on `@dinero.js/currencies` before it can run its `prepare` script and generate the currency files.

johnhooks:
@sarahdayan removing `prepare` from most packages, other than currencies works. In #667 there is still a global `prepare` script, but it is managed by `turbo`.

johnhooks:
@sarahdayan are you still interested in generating currency files?

reckter:
Is there a reason this wasn't merged in the end? Did I just not see replacement for this? 

If I just completely missed it, my bad :) 

But as  #582 is still open, we probably need a solution for this topic

PR #672, opened 2022-11-08
ci(#671): remove duplication of calling prepare

### LINKED ISSUE #671: [Bug]: build steps are run twice during CI

johnhooks:
I noticed `prepare` is automatically run after `yarn install` in every Circle CI job. By adding the flag `--ignore-scripts` it removes duplication.

Closes #671

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ aliases:
       - ~/.cache/yarn
   - &run_yarn_install
     name: Install dependencies
-    command: yarn install --frozen-lockfile
+    command: yarn install --frozen-lockfile --ignore-scripts
   - &restore_dist_folders
     name: Restore dist folders
     command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,12 +82,10 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - *attach_workspace
       - run: *install_yarn_version
       - restore_cache: *restore_yarn_cache
       - run: *run_yarn_install
       - save_cache: *save_yarn_cache
-      - run: *restore_dist_folders
       - run:
           name: Linting
           command: yarn run lint
@@ -95,10 +93,12 @@ jobs:
     <<: *defaults
     steps:
       - checkout
+      - *attach_workspace
       - run: *install_yarn_version
       - restore_cache: *restore_yarn_cache
       - run: *run_yarn_install
       - save_cache: *save_yarn_cache
+      - run: *restore_dist_folders
       - run:
           name: Type checking
           command: yarn run test:types
@@ -106,12 +106,10 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - *attach_workspace
       - run: *install_yarn_version
       - restore_cache: *restore_yarn_cache
       - run: *run_yarn_install
       - save_cache: *save_yarn_cache
-      - run: *restore_dist_folders
       - run:
           name: Unit tests
           command: yarn run test --maxWorkers=4
@@ -134,12 +132,10 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - *attach_workspace
       - run: *install_yarn_version
       - restore_cache: *restore_yarn_cache
       - run: *run_yarn_install
       - save_cache: *save_yarn_cache
-      - run: *restore_dist_folders
       - run:
           name: Release if needed
           command: yarn run shipjs trigger
@@ -149,13 +145,11 @@ workflows:
   ci:
     jobs:
       - build
-      - test_lint:
+      - test_lint
+      - test_types:
           requires:
             - build
-      - test_types
-      - test_unit:
-          requires:
-            - build
+      - test_unit
       - test_size:
           requires:
             - build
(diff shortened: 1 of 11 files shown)


johnhooks:
I have been noticing this happening a lot during the `test_size` job. It'll fail because of a missing dependency of `bundlesize`.

<details>
 <summary>Error Log</summary>

```sh
yarn run v1.22.5
$ bundlesize
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module './build/bindings/iltorb.node'
Require stack:
- /home/circleci/dinero.js/node_modules/iltorb/index.js
- /home/circleci/dinero.js/node_modules/brotli-size/index.js
- /home/circleci/dinero.js/node_modules/bundlesize/src/compressed-size.js
- /home/circleci/dinero.js/node_modules/bundlesize/src/files.js
- /home/circleci/dinero.js/node_modules/bundlesize/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/home/circleci/dinero.js/node_modules/iltorb/index.js:10:40)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/home/circleci/dinero.js/node_modules/iltorb/index.js',
    '/home/circleci/dinero.js/node_modules/brotli-size/index.js',
    '/home/circleci/dinero.js/node_modules/bundlesize/src/compressed-size.js',
    '/home/circleci/dinero.js/node_modules/bundlesize/src/files.js',
    '/home/circleci/dinero.js/node_modules/bundlesize/index.js'
  ]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Exited with code exit status 1
```

</details>

johnhooks:
I got it, a dependency of `bundlesize` is a native addon and requires running the `postinstall` script. So using `--ignore-script` probably isn't the solution. But I reverted just `test_size` back to the origin `yarn install --frozen-lockfile` command not `yarn install --frozen-lockfile --ignore-scripts`

johnhooks:
With the packages release being run at the end of the CI workflow, is the `prepare` script necessary? Otherwise I'm uncertain how to fix the duplication and insure the fix doesn't interfere with a currency or future dependancy.

sarahdayan:
@johnhooks I've added a bit more context and a potential solution [in the issue](https://github.com/dinerojs/dinero.js/issues/671#issuecomment-1311947833), let me know what you think.

johnhooks:
This PR improves the CircleCI build time from `7m 25s` to `3m 14s` an improvement of **56%**.

sarahdayan:
@johnhooks Should we keep the `prepare` scripts and use `--ignore-scripts` in the CI? This way, when working locally, you don't need to build manually to run a bundle size check or make the documentation work.

johnhooks:
@sarahdayan the problem with `--ignore-scripts` is that it doesn't run any scripts for all the dependencies on install. And specifically, `bundlesize` has a dependency that requires running a script during install. So it could be overridden for just the `test_size` job, but then it will rerun building and take 1-2 extra minutes.

> make the documentation work.

Can you point me to the issue with the documentation not working without the prepare script?

johnhooks:
Locally I was having issue with the prepare script because different branches I was working on had different devDependencies, and every time I switched branches I had to run `yarn install` and it would always run the build steps, which is mildly irritating... but definitely something I can live with if necessary.

johnhooks:
After thinking about it, I haven't worked on the docs website locally. I will run it and see what headaches this change make to that development process. The website needs to find the types and build files. The types can be fixed with some edits to the website `tsconfig.json` file. I'll think about the build artifacts and how to cleanly resolve this issue.

johnhooks:
@sarahdayan I believe #667 solves this problem in a better way.

johnhooks:
I 'm closing this in favor of using TurboRepo to solve this problem #667.

PR #673, opened 2022-11-16
fix: don't accept extra fields inside the currency options on create

fivism:
In order to save and retrieve dineros to/from our storage layer we map them to a copycat money-type. However, on retrievals we noticed that internal fields were leaking into the currency objects like so:

```
"amount": 0,
"currency": {
    "_type": "currencyObj",  <----
    "base": 10,
    "code": "USD",
    "exponent": 2
},
"scale": 2
```
... and upon further investigation one can pass anything into a dinero's currency object as long as the required subset of fields is met. This change makes the selection of fields as explicit as the base structure's (where `amount`, `currency` and `scale` are taken and everything else ignored).

### DIFF
--- a/packages/core/src/helpers/createDinero.ts
+++ b/packages/core/src/helpers/createDinero.ts
@@ -12,9 +12,11 @@ export function createDinero<TAmount>({
 }: CreateDineroOptions<TAmount>) {
   return function dinero({
     amount,
-    currency,
-    scale = currency.exponent,
+    currency: { code, base, exponent },
+    scale = exponent,
   }: DineroOptions<TAmount>): Dinero<TAmount> {
+    const currency = { code, base, exponent };
+
     onCreate?.({ amount, currency, scale });
 
     return {
--- a/packages/dinero.js/src/__tests__/dinero.test.ts
+++ b/packages/dinero.js/src/__tests__/dinero.test.ts
@@ -17,4 +17,20 @@ describe('dinero', () => {
 
     expect(snapshot).toMatchObject({ amount: 500, currency: USD, scale: 2 });
   });
+  it('cleans up unwanted properties from the options', () => {
+    const d = dinero({
+      amount: 500,
+      // @ts-expect-error
+      currency: { code: 'USD', exponent: 2, base: 10, _extraProperty: 123 },
+      _extraProperty: 123,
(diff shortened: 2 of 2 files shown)


packages/core/src/helpers/createDinero.ts
```diff
@@ -23,7 +23,11 @@ export function createDinero<TAmount>({
       toJSON() {
         return {
           amount,
-          currency,
+          currency: {
+            code: currency.code,
+            base: currency.base,
+            exponent: currency.exponent,
+          },
```
sarahdayan, line 30: This works fine for `toJSON` but `onCreate` in l. 18 still receives the full currency object.

My recommendation would be to destructure each property of `currency` in the parameters, then create a `currency` constant in the function body from these destructured properties. This way the rest of the code doesn't have to change.

Performance-wise, this should be negligible.

packages/dinero.js/src/__tests__/dinero.test.ts
```diff
@@ -17,4 +17,20 @@ describe('dinero', () => {
...
+    const d = dinero(dineroReadyObj);
+
+    const snapshot = toSnapshot(d);
+
+    expect(snapshot).toStrictEqual({
+      amount: 500,
+      currency: USD,
+      scale: 2,
+    });
+  });
```
sarahdayan, line 35:
```suggestion
  it('cleans up unwanted properties from the options', () => {
    const d = dinero({
      amount: 500,
      // @ts-expect-error
      currency: { code: 'USD', exponent: 2, base: 10, _extraProperty: 123 },
      _extraProperty: 123,
    });

    const snapshot = toSnapshot(d);

    expect(snapshot).toStrictEqual({
      amount: 500,
      currency: USD,
      scale: 2,
    });
  });
```

sarahdayan (commented):
Good one, thanks!

I left two minor comments, then we'll be good to go.

packages/core/src/helpers/createDinero.ts
fivism: Thanks and thanks for this work!

sarahdayan (approved):

sarahdayan:
Thanks for your contribution @fivism!

PR #678, opened 2022-11-23
docs: add deno example

johnhooks:
I've been playing with using Dinero.js with Deno. It requires using an import map. I thought an example might be helpful.

I will need to figure out how to change the version numbers in the `import_map.json`. I also don't know if for the example the `deno.lock` file should be committed.

The example has the exact same contents as `example/starter`.


sarahdayan:
Neat! I've pushed some slight changes to use npm specifiers, this way there's no need for import maps.

johnhooks:
@sarahdayan I forgot about the npm specifier, thanks! That is cleaner.

johnhooks:
Can codesandbox run Deno modules?

sarahdayan:
@johnhooks Not natively. We need to use a `package.json` file and install Deno (see example [here](https://codesandbox.io/s/y56n2?file=/package.json)).

I think it's fine not to do it.

sarahdayan:
One thing to figure out before moving on though is how to make this example play ball with Ship.js and Lerna.

When releasing a new version with Ship.js, once the new version is selected, Lerna runs to update the version of the library everywhere including in the examples (in the `package.json`). Then, Ship.js runs `yarn install` which updates `yarn.lock`, etc.

This won't work out of the box with the Deno example, as I don't think it supports Deno at all.

Maybe one lead is to use relative dependencies instead of npm specifiers?

johnhooks:
That works if someone downloads the whole repo, but it is confusing if a beginner uses the example as an "example" and copy and pastes.

Maybe relative imports, details in the README.md, a docs page about importing in Deno, and a preemptive GitHub issue or discussion could help solve possible confusion.

sarahdayan:
@johnhooks I hoped we could use range versions (`@2`) but it doesn't appear to be working… maybe because the packages are still alpha?

Another route would be to update [this script](https://github.com/dinerojs/dinero.js/blob/main/ship.config.js#L18-L27) to update the versions in the example and re-compile the lock file.

johnhooks:
But the lockfile won't update until the packages are published because the aren't available... right?

johnhooks:
I tried the range version with `npm:date-fns@2` and it worked. So it probably will work when `dinero.js` is out of alpha. Though it does add a specific version to the lock file. So it doesn't really fix anything. But it is a rather small issue, especially with such a simple example. 

I like the changes you made.

johnhooks:
What I was referring to was the lock file specifically. Unless you want to manually edit it, it can't be updated until after the packages are released. It can probably be cleaned up with a script, but ship.js doesn't have a `packagesPublished` hook. It looks like there is a hook for Slack, but not a regular hook.

sarahdayan:
I forgot that there's sub-resource integrity here 🤦 

Another solution is to remove the lock file and exclude it from version control. Users can build it locally. It's only an example, this isn't that big of a deal.

johnhooks:
> Another solution is to remove the lock file and exclude it from version control

I think that is simplest

johnhooks:
@sarahdayan I removed the lock file and rebase to main. It should be ready.

sarahdayan:
@johnhooks Yes let's ignore this example for now. I couldn't find anything related to it in the Deno docs, and I believe npm specifiers are still [quite experimental](https://github.com/denoland/deno/issues/15960) so let's move forward.

johnhooks:
@sarahdayan It should be good now.

sarahdayan (approved):
Go Deno 🚀

PR #679, opened 2022-11-27
build: fix build failure because of type checking tests

johnhooks:
I added an exclude directive to the `tsconfig.json` files in the packages.

@sarahdayan wIth this change, I get type errors in my editor for tests files, and the `test:types` fails if there is a type error in a test. Though it doesn't attempt to build the tests, which is why the build is currently failing.


johnhooks:
@sarahdayan with this change everything passes, but I do get type errors in my editor, which is an improvement.

sarahdayan (approved):
Thanks for fixing this!

johnhooks:
@sarahdayan I didn't know `fix:` did that, I'll remember that.

PR #682, opened 2022-11-29
build: add explicit return types

### LINKED ISSUE #681: test: ensure return type of functions
There was an odd issue in PR #309, where a function returned `unknown` rather than the expected type but it didn't raise any type errors and passed all the tests.

There must be a standard way to test types, and prevent this in the future.

johnhooks:
Add explicit return types to all functions.

fix #681


packages/core/src/types/UnaryOperation.ts
```diff
@@ -1,3 +1,4 @@
-export type UnaryOperation<TInput, TOutput = TInput> = (
-  value: TInput
+export type UnaryOperation<TInput, TParams = readonly [], TOutput = TInput> = (
+  operand: TInput,
+  ...rest: TParams extends readonly [...infer TRest] ? TRest : never
```
johnhooks, line 3: @sarahdayan what do you think of this? I would also change BinaryOperation to be simular. Without something like this adding return types to all the core functions is really messy looking and receptive.

packages/core/src/api/toUnits.ts
```diff
@@ -1,20 +1,28 @@
...
-export function toUnits<TAmount, TOutput>(calculator: Calculator<TAmount>) {
+export function toUnits<TAmount, TOutput>(
+  calculator: Calculator<TAmount>
+): UnaryOperation<
+  Dinero<TAmount>,
+  readonly [
+    transformer?: Transformer<TAmount, TOutput, readonly TAmount[]> | undefined
+  ],
+  TOutput
+> {
```
johnhooks, line 17: Complicated unary function are still going to be messy looking.

packages/core/src/api/trimScale.ts
```diff
@@ -1,18 +1,22 @@
...
 import { computeBase, countTrailingZeros, equal, maximum } from '../utils';
 
 import { transformScale } from './transformScale';
 
 export type TrimScaleParams<TAmount> = readonly [dineroObject: Dinero<TAmount>];
 
-export function trimScale<TAmount>(calculator: Calculator<TAmount>) {
+export function trimScale<TAmount>(
+  calculator: Calculator<TAmount>
+): UnaryOperation<Dinero<TAmount>> {
```
johnhooks, line 10: But simple functions have a very clean return type.

sarahdayan:
@johnhooks Yes, that's one of the items in the [v2 beta project](https://github.com/orgs/dinerojs/projects/3) (somehow I can't link directly to a card, but it's the **Reconsider package breakdown** one.

I'm not convinced we even need a `core` package anymore. It was initially done because the plan was to provide Dinero.js with two APIs (chainable and functional) but this was dropped.

Circular dependencies shouldn't be an issue since both the the currencies and calculator are dependency-free.

johnhooks:
@sarahdayan I can't comment in the project section. Though I did leave a comment in the v2 discussion. https://github.com/dinerojs/dinero.js/discussions/618#discussioncomment-4267398

johnhooks:
@sarahdayan I think `core` is important for types, especially with the TS project references, there can't be circular dependencies. If all the types are in `dinero.js` and another package needs them, but is then also imported back into `dinero.js` it will break.

johnhooks:
Also if `core` is only types, it can be listed as a devDep and won't be imported by the end user.

johnhooks:
Though if everything, but the currencies, is moved into the main package then it doesn't matter.

johnhooks:
@sarahdayan 

> Circular dependencies shouldn't be an issue since both the the currencies and calculator are dependency-free.

The calculators import core for types

sarahdayan:
>I can't comment in the project section.

I turned the project card into an issue (#683), you can comment on it now.

>If all the types are in `dinero.js` and another package needs them, but is then also imported back into `dinero.js` it will break.
>The calculators import core for types

Yup, but I'm not convinced calculators should be separate. They could be in `dinero.js` and as far as UMD is concerned, be exposed as a separate UMD endpoint.

We can discuss this all in #683 to avoid scattering thoughts.

johnhooks:
@sarahdayan I'm not going to touch this until some resolution to issues #685 and
#683. 

Do you want help accomplishing this? If not I will close this PR.

I'm definitely willing to help, but understand if you want to tackle this yourself.

sarahdayan:
Not sure why this got closed.

sarahdayan:
Looks like it got closed because #309 was merged. Normally it should have switched target branch for `main`.

Feel free to reopen @johnhooks.

johnhooks:
@sarahdayan its not giving me the option to reopen. We can open a new PR once decisions have been made about reorganizing the structure. If `core/src/api` moves to `dinero.js` it would simplify a lot of the typing.

PR #686, opened 2022-11-30
build: stop formatting api files

johnhooks:
I finally figured out why the `*.api.md` files kept being reformatted! `api-extractor` uses the code in the `lib` folder to build the api files, but it wasn't ignored by Prettier so anytime the script `format` ran it would format all the files in the lib folders, then on the next run of `build:types` the reformatted code was used!

## Changes in this PR

- Add `lib` to `.prettierignore`
- Rebuild all the `*.api.md` files to original output from `api-extractor`

Now the `*.api.md` files should only change if the api does.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,7 @@
 coverage
 node_modules
 dist
+lib
 build
 website
 **/*.md
--- a/packages/calculator-bigint/etc/calculator-bigint.api.md
+++ b/packages/calculator-bigint/etc/calculator-bigint.api.md
@@ -15,9 +15,7 @@ export const add: BinaryOperation<bigint>;
 // @public (undocumented)
 export const calculator: {
     add: BinaryOperation<bigint, bigint>;
-    compare: BinaryOperation<bigint,
-    ComparisonOperator
-    >;
+    compare: BinaryOperation<bigint, ComparisonOperator>;
     decrement: UnaryOperation<bigint, bigint>;
     increment: UnaryOperation<bigint, bigint>;
     integerDivide: BinaryOperation<bigint, bigint>;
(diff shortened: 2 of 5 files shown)


sarahdayan:
Uh! How comes [this line](https://github.com/dinerojs/dinero.js/blob/185e9e8f4e50cf884ad419cba6eebca13d703453/.prettierignore#L6) didn't work? 🤨

johnhooks:
@sarahdayan because `api-extractor` must use the declaration files as is and doesn't reformat it.

Steps for issue to happen

- `tsc` outputs to `lib`
- `api-extractor` rolls up declarations using code in lib
- code is committed 
- `lib` isn't removed because leaving it allows for incremental builds, `tsc` won't rebuild what hasn't changed. 
- later on someone runs the script `format` and the `lib` directory is formatted because it isn't excluded in `.prettierignore`
- then someone runs `build` or `build:types`
- `tsc` runs but the code in `src` hasn't changed, so it doesn't output new files to `lib`, *they are still the reformatted files.*
- `api-extractor` runs using the reformatted code, and rolls up the declaration and `*.api.md` with the *code that was reformatted by Prettier!*

Does that make sense... it's a pretty convoluted, multistep issue.

johnhooks:
@sarahdayan Prettier didn't touch the `*.api.md` files, it touched all the `lib/**/*.d.ts` that `api-extractor` used to generate the `*.api.md` files. We could always clean the `lib` folders, but we would lose incremental build, but Prettier shouldn't be touching build artifacts, otherwise the problem can happen again when build step are used individually.

johnhooks:
@sarahdayan if you want to see exactly what happens.

```sh
$ rm -rf packages/*/lib
$ yarn build:types # if your local *.api.md files have been previously modified some of them will have changed
$ git add -A # so you can see what happens next
$ yarn format # watch the output, it formats all the packages/*/lib/**/*.d.ts files, but that didn't modify the git status
$ yarn build:types # *.api.md files have now been generated with the reformatted files
```

sarahdayan:
Ah, gotcha now 👍 I was missing that it's not `.md` files right away.

sarahdayan (approved):

PR #687, opened 2022-12-02
ci: test github action

johnhooks:


johnhooks:
I made this by mistake, I meant to do it in my fork, but it defaulted to `dinerojs/dinero.js` not `johnhooks/dinero.js` sorry

PR #690, opened 2022-12-06
fix(toDecimal): teach `toDecimal` how to handle negative units

jparise:
Previously, toDecimal would format an amount like -1050 as '-10.-50', when we'd expect '-10.50'. We now use the absolute value of all but the first unit when building the decimal string.

### DIFF
--- a/packages/core/src/api/toDecimal.ts
+++ b/packages/core/src/api/toDecimal.ts
@@ -1,7 +1,7 @@
 import { NON_DECIMAL_CURRENCY_MESSAGE } from '../checks';
 import { assert } from '../helpers';
 import type { Calculator, Dinero, Formatter, Transformer } from '../types';
-import { computeBase, equal, isArray } from '../utils';
+import { absolute, computeBase, equal, isArray } from '../utils';
 
 import { toUnits } from './toUnits';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/api/toDecimal.ts
```diff
@@ -31,7 +32,9 @@ export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
     // eslint-disable-next-line functional/no-expression-statement
     assert(isDecimal, NON_DECIMAL_CURRENCY_MESSAGE);
 
-    const units = toUnitsFn(dineroObject);
+    const units = toUnitsFn(dineroObject).map((unit, index) => {
+      return index === 0 ? unit : absoluteFn(unit);
+    });
```
sarahdayan, line 37: We could do this in `getDecimal` instead, where we already map over units. This also gathers all the decimal-related transforms in one place.

sarahdayan (commented):
Great catch, thank you!

One small change and we'll be good to go.

packages/core/src/api/toDecimal.ts
jparise: I started with that approach, but then I realized I'd need to supply `getDecimal` with the `absoluteFn` and then took the above approach instead.

... but I agree it would be nicer to consolidate the transformation, so I'll adjust.

sarahdayan (approved):
LGTM 🚀

PR #693, opened 2022-12-10
fix(toDecimal): preserve negative sign for leading zeros

### LINKED ISSUE #692: [Bug]: toDecimal fails to print a leading negative sign when the first unit is negative zero

### Current behavior

#690 fixed the the general case of formatting negative unit values, but there is one outstanding specific case where formatting is still incorrect: when the first unit is -0, the resulting string doesn't include the leading negative sign.

This occurs because the (default) `toString` formatter is `String`, and `String(-0)` produces `'0'`.

```ts
toDecimal(dinero({ amount: -1, currency: USD }) === '0.01'
```

### Expected behavior

```ts
toDecimal(dinero({ amount: -1, currency: USD }) === '-0.01'
```

### Steps to reproduce

Here's a failing test case:

```ts
it('returns the negative amount for the last unit in decimal format', () => {
  const d = dinero({ amount: -1, currency: USD });

  expect(toDecimal(d)).toEqual('-0.01');
});
```
```
Expected: "-0.01"
Received: "0.01"
```

### Version

2.0.0-alpha12

### Environment

Node.js 16.14.1

### Code of Conduct

jparise:
Change #690 fixed the general case of formatting negative unit values, but was one remaining case where formatting was still incorrect: when the first unit is -0, the resulting string didn't include the leading negative sign.

This change identifies that exact case (negative value, leading zero) and pads the resulting string with a leading negative sign.

Fixes #692

### DIFF
--- a/packages/core/src/api/toDecimal.ts
+++ b/packages/core/src/api/toDecimal.ts
@@ -54,17 +54,17 @@ function getDecimal<TAmount>(
   const zero = calculator.zero();
 
   return (units: readonly TAmount[], scale: TAmount) => {
-    const integer = formatter.toString(units[0]);
+    const whole = formatter.toString(units[0]);
     const fractional = formatter.toString(absoluteFn(units[1]));
 
     const scaleNumber = formatter.toNumber(scale);
-    const decimal = `${integer}.${fractional.padStart(scaleNumber, '0')}`;
+    const decimal = `${whole}.${fractional.padStart(scaleNumber, '0')}`;
 
-    // A leading negative zero is a special case because the toString
-    // formatter (generally String) won't preserve its negative sign.
     const leadsWithZero = equalFn(units[0], zero);
     const isNegative = lessThanFn(units[1], zero);
 
+    // A leading negative zero is a special case because the `toString`
+    // formatter won't preserve its negative sign (since 0 === -0).
     return leadsWithZero && isNegative ? `-${decimal}` : decimal;
   };
 }
--- a/packages/core/src/api/toDecimal.ts
+++ b/packages/core/src/api/toDecimal.ts
@@ -1,7 +1,7 @@
 import { NON_DECIMAL_CURRENCY_MESSAGE } from '../checks';
 import { assert } from '../helpers';
 import type { Calculator, Dinero, Formatter, Transformer } from '../types';
-import { absolute, computeBase, equal, isArray } from '../utils';
+import { absolute, computeBase, equal, isArray, lessThan } from '../utils';
 
 import { toUnits } from './toUnits';
 
@@ -49,23 +49,22 @@ function getDecimal<TAmount>(
   formatter: Formatter<TAmount>
 ) {
   const absoluteFn = absolute(calculator);
+  const equalFn = equal(calculator);
+  const lessThanFn = lessThan(calculator);
+  const zero = calculator.zero();
 
   return (units: readonly TAmount[], scale: TAmount) => {
-    return units
-      .map((unit, index) => {
-        const isFirst = index === 0;
-        const isLast = units.length - 1 === index;
+    const integer = formatter.toString(units[0]);
+    const fractional = formatter.toString(absoluteFn(units[1]));
 
-        const unitAsString = formatter.toString(
-          isFirst ? unit : absoluteFn(unit)
-        );
+    const scaleNumber = formatter.toNumber(scale);
+    const decimal = `${integer}.${fractional.padStart(scaleNumber, '0')}`;
 
-        if (isLast) {
-          return unitAsString.padStart(formatter.toNumber(scale), '0');
-        }
+    // A leading negative zero is a special case because the toString
+    // formatter (generally String) won't preserve its negative sign.
+    const leadsWithZero = equalFn(units[0], zero);
+    const isNegative = lessThanFn(units[1], zero);
 
-        return unitAsString;
-      })
-      .join('.');
+    return leadsWithZero && isNegative ? `-${decimal}` : decimal;
   };
 }
--- a/packages/dinero.js/src/api/__tests__/toDecimal.test.ts
+++ b/packages/dinero.js/src/api/__tests__/toDecimal.test.ts
@@ -45,6 +45,16 @@ describe('toDecimal', () => {
 
         expect(toDecimal(d)).toEqual('-10.50');
       });
+      it('returns the negative amount with a leading zero in decimal format', () => {
+        const d = dinero({ amount: -1, currency: USD });
+
+        expect(toDecimal(d)).toEqual('-0.01');
+      });
+      it('returns negative zero amount as a positive value in decimal format', () => {
+        const d = dinero({ amount: -0, currency: USD });
+
+        expect(toDecimal(d)).toEqual('0.00');
+      });
       it('uses a custom transformer', () => {
         const d = dinero({ amount: 1050, currency: USD });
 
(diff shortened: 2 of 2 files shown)


packages/core/src/api/toDecimal.ts
```diff
@@ -32,9 +41,15 @@ export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
...
+    const isNegative = units.some((unit) => lessThanFn(signFn(unit), zero));
 
     const getDecimalFn = getDecimal(calculator, dineroObject.formatter);
-    const value = getDecimalFn(units, scale);
+    const decimalValue = getDecimalFn(units, scale);
+
+    // A leading negative zero is a special case because the toString
+    // formatter (generally String) won't preserve its negative sign.
+    const value =
+      isNegative && equalFn(units[0], zero) ? `-${decimalValue}` : decimalValue;
```
jparise, line 52: I could also use `isZero` here, but we already have all of the components at hand to perform this check. Let me know if `isZero` is preferred.

packages/core/src/api/toDecimal.ts
```diff
@@ -32,9 +41,15 @@ export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
     assert(isDecimal, NON_DECIMAL_CURRENCY_MESSAGE);
 
     const units = toUnitsFn(dineroObject);
+    const isNegative = units.some((unit) => lessThanFn(signFn(unit), zero));
```
sarahdayan, line 44: We should be able to simplify this further and not even have to map over units (either here or even reusing the existing `Array.map` in `getDecimal`).

Normally, when using `toUnits` with a decimal Dinero objects (base 10 with a single base) it should always return a tuple of 2.

```js
toUnits(dinero({ amount: -1, currency: USD })) // [-0, -1]
toUnits(dinero({ amount: -0, currency: USD })) // [-0, -0]
```

If the first unit is zero, then we can just test the second unit for being negative instead of mapping over all units. This way, the second test only happens when necessary (when the first unit is zero) and we avoid an unnecessary iteration.

packages/core/src/api/toDecimal.ts
sarahdayan: Importing `isZero` might make `toDecimal` slightly heavier, so I prefer your approach.

packages/core/src/api/toDecimal.ts
sarahdayan, line 44: Not sure you need `sign`. Comparing `unit` against `zero` should already work.

packages/core/src/api/toDecimal.ts
```diff
@@ -32,9 +41,15 @@ export function toDecimal<TAmount, TOutput>(calculator: Calculator<TAmount>) {
...
 
     const getDecimalFn = getDecimal(calculator, dineroObject.formatter);
-    const value = getDecimalFn(units, scale);
+    const decimalValue = getDecimalFn(units, scale);
```
sarahdayan, line 47: I would probably perform the two checks (whether it leads with zero and is negative) inside `getDecimal`—I expect this function to return the correct value. wdyt?

packages/core/src/api/toDecimal.ts
jparise: Yes, that makes good sense. Can we always assume there's a second value in the tuple? I used `units.length - 1` just in case that's not always the case.

jparise: Yes, this looks pretty funny in hindsight! I started with `sign`, then realized it returns a `TAmount` value, and then introduced the `lessThan` check to handle it. 😄

packages/core/src/api/toDecimal.ts
```diff
@@ -49,12 +49,17 @@ function getDecimal<TAmount>(
...
 
   return (units: readonly TAmount[], scale: TAmount) => {
-    return units
+    const lastIndex = units.length - 1;
+
+    const decimalString = units
       .map((unit, index) => {
         const isFirst = index === 0;
-        const isLast = units.length - 1 === index;
+        const isLast = lastIndex === index;
```
jparise, line 62: I think we could also drop `isFirst` and `isLast` in favor of comparing directly to `index` in the following lines. `index === 0` and `index === lastIndex` are pretty self-descriptive. I don't know if using constants is more v8-friendly, though.

packages/core/src/api/toDecimal.ts
sarahdayan: When wasn't it the case? With decimal objects it should be.

packages/core/src/api/toDecimal.ts
sarahdayan: I'm fine with inlining as long as it doesn't hurt readability, as the performance gains are usually very small.

If you think that works well here, go for it.

jparise:
Given that `getDecimal` always receives a two-element `units` array, I took a more "drastic" approach and unrolled the `.map()` operation. We were specializing each element (first, last) within the mapping function anyway.

(`getDecimal` could `assert(units.length === 2, ...)`, but I think the `assert()` in `toDecimal` already covers this "must be decimal" case?)

I think this results in a simpler solution overall that only uses a single top-level branch (to determine if we need to add the leading negative).

packages/core/src/api/toDecimal.ts
```diff
@@ -49,23 +49,22 @@ function getDecimal<TAmount>(
...
-        );
-
-        if (isLast) {
-          return unitAsString.padStart(formatter.toNumber(scale), '0');
-        }
-
-        return unitAsString;
-      })
-      .join('.');
+    const integer = formatter.toString(units[0]);
```
sarahdayan, line 57:
```suggestion
    const whole = formatter.toString(units[0]);
```

packages/core/src/api/toDecimal.ts
```diff
@@ -49,23 +49,22 @@ function getDecimal<TAmount>(
...
+    const fractional = formatter.toString(absoluteFn(units[1]));
+
+    const scaleNumber = formatter.toNumber(scale);
+    const decimal = `${integer}.${fractional.padStart(scaleNumber, '0')}`;
```
sarahdayan, line 61:
```suggestion
    const decimal = `${whole}.${fractional.padStart(scaleNumber, '0')}`;
```

packages/core/src/api/toDecimal.ts
```diff
@@ -49,23 +49,22 @@ function getDecimal<TAmount>(
...
+
+    // A leading negative zero is a special case because the toString
+    // formatter (generally String) won't preserve its negative sign.
```
sarahdayan, line 64:
```suggestion
```

packages/core/src/api/toDecimal.ts
```diff
@@ -49,23 +49,22 @@ function getDecimal<TAmount>(
...
+    const leadsWithZero = equalFn(units[0], zero);
+    const isNegative = lessThanFn(units[1], zero);
+
+    return leadsWithZero && isNegative ? `-${decimal}` : decimal;
```
sarahdayan, line 68:
```suggestion
    // A leading negative zero is a special case because the `toString`
    // formatter won't preserve its negative sign (since 0 === -0).
    return leadsWithZero && isNegative ? `-${decimal}` : decimal;
```

sarahdayan (commented):
Excellent rewrite! A few suggestions and we'll be good to go.

sarahdayan (approved):
Awesome contribution, thanks @jparise!

PR #694, opened 2022-12-11
refactor: move constants into closures

johnhooks:
I noticed that not all of the constants were included in the closures of the core functions. There is no reason to recalculate these values because they will never change.

### DIFF
--- a/packages/core/src/api/hasSubUnits.ts
+++ b/packages/core/src/api/hasSubUnits.ts
@@ -8,6 +8,7 @@ export type HasSubUnitsParams<TAmount> = readonly [
 export function hasSubUnits<TAmount>(calculator: Calculator<TAmount>) {
   const equalFn = equal(calculator);
   const computeBaseFn = computeBase(calculator);
+  const zero = calculator.zero();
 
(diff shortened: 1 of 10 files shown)


johnhooks:
@sarahdayan I'm looking at this closer... are all the core functions are recreated on every call of an api function in `dinero.js`?

sarahdayan:
Thanks for looking into it.

Yes, core functions are recreated on every function call, that's a trade-off of using independent, tree-shakeable functions. I don't think this is a bad trade-off TBH because creating these functions isn't _that_ costly. It isn't as CPU-optimal but it's more memory effective.

Regarding your refactor, if you look into each file you'll notice that the functions you hoisted were all created after early returns. I purposely created them after to save up on CPU and memory. No need to create them if you never use them, which you only know at runtime.

As you pointed it out, the core function is always recreated so right now the refactor should have no impact on performance.

If you want to look deeper into it you're much welcome! However I'm not convinced this is where you'll reap the best performance gains—at least, even if it is, this is unlikely to make a noticeable difference. I'd recommend profiling an actual app that uses Dinero.js (maybe one of the examples?) to identify what really slows it down.

PR #696, opened 2022-12-11
refactor(core): add createCalculator

johnhooks:
By creating a more complicated calculator object in `createDinero`, core functions could be optimized by not recreating constants in closures on every function call in the `dinero.js` api. It's based off some observations in #694.

@sarahdayan this is just an idea with some sudo code that isn't complete. Let me know what you think.


johnhooks:
Scratch that idea, I just realized that ruins the tree shaking. I'll keep thinking on it, there should be a way to simplify it.

PR #699, opened 2022-12-12
build: refine project references

johnhooks:
After reading an article from Shopify and how they implemented ts project references, I learned a few things I wanted to apply. It should specifically speed up VSCode, and possibly any program using the ts lsp. [reference](https://shopify.engineering/migrating-large-typescript-codebases-project-references)

The big suggestion from the article was that the top level `tsconfig.json` should be able to build the whole project, thats how VSCode will use it. I moved everything shared to a `tsconfig.base.json` and removed the `tsconfig.declaration.json` because that didn't have a purpose with a shared base between root level and leaf level `tsconfig.json` files. It also required a `tsconfig.json` inside the `packages` directory, and some modifications to the one in the `test` directory.


.eslintrc.js
```diff
@@ -47,13 +47,15 @@ module.exports = {
...
         'import/no-extraneous-dependencies': 'off',
+        'import/no-named-as-default': 'off',
       },
     },
     {
       files: ['**/__tests__/**'],
       rules: {
         'functional/no-expression-statement': ['off'],
         'import/no-extraneous-dependencies': ['off'],
+        'import/no-named-as-default': 'off',
```
johnhooks, line 58: Fixing the project references and including `project: ['./']` in `.eslint.js` (so it could find the tsconfig paths) must have let the `eslint-plugin-import` better analyze the import statements, because it started failing this import.

```js
import Big from 'big.js`;
```

The rule `import/no-named-as-default` prefers:

```js
import { Big } from 'big.js';
```

So we can change it or ignore it.

package.json
```diff
@@ -61,7 +61,7 @@
     "eslint-config-algolia": "20.0.0",
     "eslint-config-prettier": "8.5.0",
     "eslint-import-resolver-alias": "1.1.2",
-    "eslint-import-resolver-typescript": "2.7.1",
+    "eslint-import-resolver-typescript": "3.5.2",
```
johnhooks, line 64: Updating this dep to use `project` config.

website/pages/docs/[[...slug]].tsx
```diff
@@ -128,7 +128,10 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
   const headings = getHeadings(source);
 
   const root = path.join(process.cwd(), '..', 'packages');
-  const packages = fs.readdirSync(root).filter(junk.not);
+  const packages = fs
+    .readdirSync(root)
+    .filter(junk.not)
+    .filter((file) => !/tsconfig\.json/.test(file));
```
Comment (johnhooks) [inline website/pages/docs/[[...slug]].tsx:134]: @sarahdayan this was a pretty odd error I ran into by adding a `tsconfig.json` to the packages directory. It was being loaded by the `BundleSize.tsx` file by mistake. This regex test isn't very elegant but works. Another solution would be to check where or not its a directory.

package.json
johnhooks: I did look at the `CHANGELOG.md` for the package, and our configuration didn't use any of the depreciated features.

.eslintrc.js
sarahdayan: I'd prefer fixing the imports and keeping the rule on. If there are imports that can't be fixed, we can ignore the rule locally.

test/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
```
sarahdayan, line 1: What is this configuration file used for?

website/pages/docs/[[...slug]].tsx
Comment (sarahdayan) [reply website/pages/docs/[[...slug]].tsx:134]: That'll do for now 👍

.eslintrc.js
johnhooks: I will change it.

test/tsconfig.json
johnhooks: It defined the folder as its own project that can be referenced by others. It prevents it from being rebuild if nothing changes. Otherwise its rebuild on every change. And by rebuild, I mean rebuild by VSCode, because it's not typically build by `tsc`. Otherwise a change in any ts file of the whole project cause VSCode to have to rebuild this folder too, which slows down Intellisense. Does that make sense?

tsconfig.base.json
```diff
@@ -0,0 +1,35 @@
...
+      "dinero.js": ["./packages/dinero.js/src/index"],
+      "@dinero.js/core": ["./packages/core/src/index"],
+      "@dinero.js/currencies": ["./packages/currencies/src/index"],
+      "@dinero.js/calculator-bigint": [
+        "./packages/calculator-bigint/src/index"
+      ],
+      "@dinero.js/calculator-number": [
+        "./packages/calculator-number/src/index"
+      ],
+      "test-utils": ["./test/utils/index"]
```
sarahdayan, line 32: One thing we do in [`algolia/instantsearch.js`](https://github.com/algolia/instantsearch.js) is use prefixes for internal dependencies (private packages and references) to avoid any issue with existing packages of the same name and make it more obvious when reading the code that this is internal.

What do you think we rename `test-utils` to `@dinero.js/test`?

test/tsconfig.json
sarahdayan: That makes perfect sense, thanks for explaining 👍

sarahdayan (commented):
Overall it looks great to me. I'll let you handle the final comments @johnhooks and we'll be good to merge.

tsconfig.base.json
johnhooks: I can definitely do that, but that was the original original alias I made 😄

https://github.com/dinerojs/dinero.js/pull/666#discussion_r1015837973

johnhooks: Do you want to just make it an external package?

Projects commonly do that with their base tsconf.json too. So it's easier to reference rather than long relative paths.

sarahdayan: HAHAHA that's a Spiderman meme situation for sure 😆

Let's keep it as it is and change that in a separate PR if we still want to.

.eslintrc.js
sarahdayan: @johnhooks This should be the last change before we can merge.

johnhooks:
@sarahdayan I made one last edit to `packages/dinero.js/src/api/__tests__/config.json`. Importing `test-utils` was causing an ESLint `import/no-unresolved` error. I think some of the `configOptions` in package level config caused the problem. I extended `tsconfig.base.json` in the root directory and everything works correctly.

Also, I looked back through the comments and it looked like everything was resolved. Though if I missed something just let me know.

sarahdayan:
@johnhooks Not seeing any issue locally, and the build seems fine.

johnhooks:
Thanks I'll start trying to figure out what is wrong on my system and stop looking in the code.

.eslintrc.js
johnhooks: I don't know why I periodically had this error happening. But I'll look locally for the solution because it isn't in CI.

johnhooks:
@sarahdayan I think it's ready now. There are some changes that need to happen in #667 for this to work. Once this is merged, I'll resolve them.

.eslintrc.js
johnhooks: It is hard jumping between this and #667, I hadn't removed the rule as you suggested. THAT made the error come back lol.

.eslintrc.js
```diff
@@ -47,15 +47,13 @@ module.exports = {
...
         'import/no-extraneous-dependencies': 'off',
-        'import/no-named-as-default': 'off',
       },
     },
     {
       files: ['**/__tests__/**'],
       rules: {
         'functional/no-expression-statement': ['off'],
         'import/no-extraneous-dependencies': ['off'],
-        'import/no-named-as-default': 'off',
```
johnhooks: @sarahdayan with was the last edit from a previous comment. To fix this ESLint issue rather than ignore it.

johnhooks:
@sarahdayan I'm having some trouble with getting ESLint to recognize the `paths` property in  `tsconfig.base.json`. From what I've learned, `tsc` can compile everything in the current mono repo structure, though it isn't the ideal setup. Ideally the `tsconfig.json` property `rootDir` would be the top of the mono repo, but that complicates outputting the files correctly into the `packages/*/lib` because it will add path of the `rootDir` to the build output location, like `packages/*/lib/packages/*/lib/`. There are some hacky solutions I have found, but I don't like any of them. 

The simplest solution would be to create a separate `packages/test-utils` and then import it though the name `@dinero.js/test-utils`. This would import in ESLint okay because Yarn would have linked the package into `node_modules` and ESLint can easily find it.

johnhooks:
@sarahdayan I'm going to rework this. Do you want me to close it and open it back up when it's ready?

sarahdayan:
@johnhooks Yes, feel free to close and reopen when you're ready so we can keep the PR list clear 🙂

PR #704, opened 2022-12-31
build: add `.next` directory to Turborepo build outputs

johnhooks:
@sarahdayan turbo wasn't caching the `.next` directory. So when it tried to use the cache it would fail. I think this should do it!

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -31,7 +31,7 @@
     },
     "build": {
       "dependsOn": ["^build", "build:esm", "build:umd:cjs", "build:types"],
-      "outputs": ["dist/**"]
+      "outputs": ["dist/**", ".next/**/*"]
     },
     "lint": {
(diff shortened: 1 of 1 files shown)


turbo.json
```diff
@@ -31,7 +31,7 @@
     },
     "build": {
       "dependsOn": ["^build", "build:esm", "build:umd:cjs", "build:types"],
-      "outputs": ["dist/**"]
+      "outputs": ["dist/**", ".next/**/*"]
```
sarahdayan, line 34: This should work the same.

```suggestion
      "outputs": ["dist/**", ".next/**"]
```

sarahdayan (dismissed):
Nice finding! I did indeed just notice it in the Turborepo docs.

sarahdayan (approved):

PR #711, opened 2023-01-11
fix(up, down, halfUp): fix handling of numbers close to 0 and rounding of integer results

### LINKED ISSUE #710: [Bug]: halfEven does not round correctly when rounding to 0

### Current behavior

halfEven does not round correctly when rounding to 0

```
const dHighPrecision = dinero({ amount: 0, currency: USD, scale: 7 });
const dUnit = transformScale(dHighPrecision, 2, halfEven);

console.log("transformed", toSnapshot(dUnit));
// transformed {amount: 1, currency: Object, scale: 2}

console.log("0/1000", halfEven(0, 1000, calculator));
// 0/1000 1
console.log("0/100", halfEven(0, 100, calculator));
// 0/100 1
console.log("0/10", halfEven(0, 10, calculator));
// 0/10 1
console.log("1/10", halfEven(1, 10, calculator));
// 1/10 1
console.log("5/10", halfEven(5, 10, calculator));
// 5/10 0 (this is correct)
console.log("51/100", halfEven(51, 100, calculator));
// 51/100 1 (this is correct)
```

### Expected behavior

Expected to see the following printed to console.
```
transformed {amount: 0, currency: Object, scale: 2}
0/1000 0
0/100 0
0/10 0
1/10 0
5/10 0
51/100 1
```

### Steps to reproduce

See

https://codesandbox.io/s/nervous-raman-q7xiyt?file=/main.js
https://q7xiyt.sse.codesandbox.io/

### Version

2.0.0-alpha13

### Environment

Windows, Chrome - Version 108.0.5359.125 (Official Build) (64-bit)

### Code of Conduct

### LINKED ISSUE #713: [Bug]: Multiplied negative values increases amount

### Current behavior

After multiplying negative numbers it adds an extra amount into the Dinero object. 

```
const d = dinero({ amount: -200, currency: USD });    // { amount: 200 }  -> OK!
const m = multiply(d, 1);                             // { amount: -201 } -> not OK!
```

Happy to share further details.

Reproducible in the attached Code Sandbox.

### Expected behavior

To be the same, but just in a negative value.

### Steps to reproduce

https://codesandbox.io/s/fervent-violet-qkozxe?file=/main.js

### Version

2.0.0-alpha.13

### Environment

Node.js v14.18.0

### Code of Conduct

shasderias:
up/down - Functions should not round integer results.

up - up should only round positive numbers, 0 is not a positive number.

halfUp - halfUp was incorrectly rounding numbers in the [0,1) range due to the incorrect definition of isPositive.

Fixes #710
Fixes #713

Written by throwing tests at the wall and making them pass, would appreciate if someone could review their correctness.

In the rounding functions, is it necessary to handle the case where factor is negative? If so, the definition of isPositive in some of the functions may not handle this case.

From a big picture going to beta POV, some of these rounding functions are really obscure (e.g. halfOdd) and have little practical application. They also feel non-trivial to get right, do we really need all of them?

Also, rounding to -0 is correct but surprising, should something be done here?

### DIFF
--- a/packages/core/src/divide/halfUp.ts
+++ b/packages/core/src/divide/halfUp.ts
@@ -1,5 +1,5 @@
 import type { DivideOperation } from '..';
-import { greaterThan, isHalf, absolute } from '../utils';
+import { absolute, greaterThan, isHalf } from '../utils';
 
 import { down, up } from '.';
 
@@ -12,12 +12,20 @@ export const halfUp: DivideOperation = (amount, factor, calculator) => {
   const remainder = absoluteFn(calculator.modulo(amount, factor));
   const difference = calculator.subtract(factor, remainder);
   const isLessThanHalf = greaterThanFn(difference, remainder);
-  const isPositive = greaterThanFn(amount, calculator.increment(zero));
+  const isPositive = greaterThanFn(amount, zero);
+
+  // Round up in these three cases:
+  // (1) the quotient is half:
+  //   e.g. ... -2.5,  -1.5,  -0.5,  0.5,  1.5,  2.5 ...
+  // (2) the quotient is positive, and greater than half:
+  //   e.g.                          0.51, 1.51, 2.51 ...
+  // (3) the quotient is negative, and less than half:
+  //   e.g. ... -2.49, -1.49, -0.49
 
   if (
     isHalfFn(amount, factor) ||
-    (isLessThanHalf && !isPositive) ||
-    (!isLessThanHalf && isPositive)
+    (isPositive && !isLessThanHalf) ||
+    (!isPositive && isLessThanHalf)
   ) {
     return up(amount, factor, calculator);
   }
--- a/packages/core/src/divide/halfUp.ts
+++ b/packages/core/src/divide/halfUp.ts
@@ -3,6 +3,21 @@ import { absolute, greaterThan, isHalf } from '../utils';
 
 import { down, up } from '.';
 
+/**
+ * Divide and round towards "nearest neighbor" unless both neighbors are
+ * equidistant, in which case round up.
+ *
+ * Rounding up happens when:
+ * - The quotient is half (e.g., -1.5, 1.5).
+ * - The quotient is positive and greater than half (e.g., 1.6).
+ * - The quotient is negative and less than half (e.g., -1.4).
+ *
+ * @param amount - The amount to divide.
+ * @param factor - The factor to divide by.
+ * @param calculator - The calculator to use.
+ *
+ * @returns The rounded amount.
+ */
 export const halfUp: DivideOperation = (amount, factor, calculator) => {
   const greaterThanFn = greaterThan(calculator);
   const isHalfFn = isHalf(calculator);
@@ -13,14 +28,6 @@ export const halfUp: DivideOperation = (amount, factor, calculator) => {
   const difference = calculator.subtract(factor, remainder);
   const isLessThanHalf = greaterThanFn(difference, remainder);
   const isPositive = greaterThanFn(amount, zero);
-
-  // Round up in these three cases:
-  // (1) the quotient is half:
-  //   e.g. ... -2.5,  -1.5,  -0.5,  0.5,  1.5,  2.5 ...
-  // (2) the quotient is positive, and greater than half:
-  //   e.g.                          0.51, 1.51, 2.51 ...
-  // (3) the quotient is negative, and less than half:
-  //   e.g. ... -2.49, -1.49, -0.49
 
   if (
     isHalfFn(amount, factor) ||
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "eslint-plugin-prettier": "4.2.1",
     "eslint-plugin-promise": "6.1.1",
     "eslint-plugin-sonarjs": "0.16.0",
+    "fast-check": "3.6.3",
     "husky": "8.0.1",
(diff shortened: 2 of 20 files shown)


johnhooks:
@shasderias this PR fixes #713 too. I added failing tests to confirm the fix. I'll add a review of this PR too.

packages/core/src/divide/halfUp.ts
```diff
@@ -12,12 +12,20 @@ export const halfUp: DivideOperation = (amount, factor, calculator) => {
...
+  //   e.g.                          0.51, 1.51, 2.51 ...
+  // (3) the quotient is negative, and less than half:
+  //   e.g. ... -2.49, -1.49, -0.49
 
   if (
     isHalfFn(amount, factor) ||
-    (isLessThanHalf && !isPositive) ||
-    (!isLessThanHalf && isPositive)
+    (isPositive && !isLessThanHalf) ||
+    (!isPositive && isLessThanHalf)
```
johnhooks, line 28: Why are these flipped? All the tests pass if I leave them as they were.

packages/core/src/divide/halfUp.ts
```diff
@@ -1,5 +1,5 @@
 import type { DivideOperation } from '..';
-import { greaterThan, isHalf, absolute } from '../utils';
+import { absolute, greaterThan, isHalf } from '../utils';
```
johnhooks, line 2: Is this edit necessary?

johnhooks (commented):
This is awesome work, you were very thorough. 

I'm reading up on half even and half odd rounding, of the new tests, I had a little trouble wrapping my head around if they are being properly tested.

packages/core/src/divide/halfUp.ts
shasderias: To be consistent with preceding comment explaining what I'm doing.

I also reckon humans handle positive numbers better, so this can be viewed as: (1) addressing the problem, then (2) the common case then (3) the less common case.

I don't feel strongly about this, except that the comment should be amended if this change is reverted.

johnhooks: That makes sense to me, I was just checking that there was a reason. It wasn't apparent.

packages/core/src/divide/halfUp.ts
shasderias: Every other import in every other file I've touched is sorted in alphabetical order except this. XD

Don't feel strongly about it though. Should probably be added as a lint if people care. My understanding is that the order of imports from different files matters in JS but not multiple imports from the same file - if this is not correct, then the change should be reverted.

shasderias:
> This is awesome work, you were very thorough.
> 
> I'm reading up on half even and half odd rounding, of the new tests, I had a little trouble wrapping my head around if they are being properly tested.

My internal logic is:
(1) if the last digit is not 5, we're not dealing with half, round to closest integer
(2) 0 is even, 1 is odd, -1 is also odd

packages/core/src/divide/halfUp.ts
johnhooks: Understood, I would agree if alphabetical order is important we should add an eslint rule for it.

johnhooks:
@shasderias I understand it better now, when I took a look at the edits in context of the existing `halfOdd` and `halfEven` tests. The existing tests already checked for rounding toward odd or even, the tests you added were to make sure it handles rounding a single base unit in the expected way.

sarahdayan:
Fantastic work @shasderias, thanks for the contirbution! I took the liberty to add a few commits instead of making suggestions, to avoid unnecessary back-and-forths. My changes are mostly refactors:

- Move your comment to JSDoc, which has the nice sife-effect of documenting the rounding functions.
- Use property tests instead of fixed test cases for when we want to work with arbitrary values that follow certain conditions.

sarahdayan:
Fantastic work @shasderias, thanks for the contirbution! I took the liberty to add a few commits instead of making suggestions, to avoid unnecessary back-and-forths. My changes are mostly refactors:

- Move your comment to JSDoc, which has the nice sife-effect of documenting the rounding functions.
- Use property tests instead of fixed test cases for when we want to work with arbitrary values that follow certain conditions.

sarahdayan:
Fantastic work @shasderias, thanks for the contirbution! I took the liberty to add a few commits instead of making suggestions, to avoid unnecessary back-and-forths. My changes are mostly refactors:

- Move your comment to JSDoc, which has the nice sife-effect of documenting the rounding functions.
- Use property tests instead of fixed test cases for when we want to work with arbitrary values that follow certain conditions.

sarahdayan:
Fantastic work @shasderias, thanks for the contirbution! I took the liberty to add a few commits instead of making suggestions, to avoid unnecessary back-and-forths. My changes are mostly refactors:

- Move your comment to JSDoc, which has the nice sife-effect of documenting the rounding functions.
- Use property tests instead of fixed test cases for when we want to work with arbitrary values that follow certain conditions.

sarahdayan:
Fantastic work @shasderias, thanks for the contirbution! I took the liberty to add a few commits instead of making suggestions, to avoid unnecessary back-and-forths. My changes are mostly refactors:

- Move your comment to JSDoc, which has the nice sife-effect of documenting the rounding functions.
- Use property tests instead of fixed test cases for when we want to work with arbitrary values that follow certain conditions.

sarahdayan (approved):

Issue #716, opened 2023-01-26
[Bug]: formatting big.js objects with scale larger than 21 fails


benknab:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I'm using big.js with dinero because I need it for cryptocurrencies and large USD values but after following the documentation to create a custom calculator I noticed that formatting breaks if the scale is too large (>= 22)

My first problem was that I receive arbitrary precision USD values from an API and I need to calculate the scale of these values. This is how I did it, but I feel I'm missing something and this has to be an overkill:

```ts
const usd = (dollars: string) => {
	// Get cents from dollars
	const cents = new Big(dollars).mul(conversion);
	// Calculate remaining fraction
	const fraction = cents.mod(1);
	const fractionLength = fraction.eq(0) ? 0 : fraction.c.length;
	// Set scale based on fraction length
	const scale = USD.exponent.plus(fractionLength);
	// Make sure integer is passed to dinero
	const amount = new Big(cents.mul(USD.base.pow(fractionLength).toFixed()));
	return dinero({
		currency: USD,
		amount,
		scale,
	});
};
```

and when I receive a value that has precision 22 or larger, `toDecimal` function breaks and the output is in scientific notation with multiple decimal points:
```ts
const n = usd("1000000000.9876543211211111111111")
toDecimal(n) // 1000000000.9.876543211211111111111e+21
```

### Expected behavior

I had to create a custom formatter to make it work:

```ts
const formatter: Formatter<Big> = {
	toNumber: (v) => (v ? v.toNumber() : 0),
	toString: (v) => (v ? v.toFixed() : ''),
};
```

Unless I'm doing something incredibly stupid with parsing the arbitrary precision, I think this should be added to the [docs](https://v2.dinerojs.com/docs/guides/using-different-amount-types#implementing-a-custom-calculator). I can open a PR if you can point me in the right direction.

And a quick question, why is the `amount` arg in the formatter functions optional?

### Steps to reproduce

https://replit.com/@bkiac/dinerowithbig#index.ts

### Version

2.0.0-alpha.13

### Environment

any

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

johnhooks:
[...]

PR #724, opened 2023-02-18
test(toDecimal): issue Dinero<Big> objects with large scale

johnhooks:
This pull request resolves issue #716. The the default `formatter` from `createDinero` didn't stringify `Big.js` values with a large scale correctly.

There is still a question about the function signatures of the `Formatter`. I created PR #725 to discuss and correct this issue.

### DIFF
--- a/packages/dinero.js/src/api/__tests__/toDecimal.test.ts
+++ b/packages/dinero.js/src/api/__tests__/toDecimal.test.ts
@@ -1,5 +1,5 @@
 import { MGA, USD } from '@dinero.js/currencies';
-import Big from 'big.js';
+import { Big } from 'big.js';
 import {
   castToBigintCurrency,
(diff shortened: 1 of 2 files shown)


johnhooks:
@sarahdayan the issue is right here. I thought a screenshot might be easier to share. Look at the value of `fractional` in the Variables on the right. I'm not very familar with `Big.js`, but the `formatter.toString` is not returning the value expected, which would be the number formatted as an integer.

Screen Shot 2023-02-18 at 11 24 50 AM

johnhooks:
Because of the high possibility of the default `formatter` argument to be the incorrect option when someone implements their own `dinero` function with `createDinero`, the formatter shouldn't be optional or there be a default argument. It would prevent someone from having this issue in the future.

